Issue #42, opened 2014-12-17
Automated child table triggers (WAS: Inherit DELETE/UPDATE triggers from parent)


davidblewett:
Currently, DELETE and UPDATE triggers that are defined on the parent will be silently ignored if they trigger action in a child table. INSERT triggers fire correctly, but from what I can tell, DELETE/UPDATE do not at all. I'm trying to automate updating an external system on changes to any data in the child tables. I can create them manually, but it would be great if pg_partman did it when it created new partitions.

keithf4:
See if RULES can be handled here as well. See https://github.com/keithf4/pg_partman/issues/121

1803:
**Trigger inheritance:**

_This is still WIP._

Side note: to make sure the `before insert` partitioning trigger fires last (we don't want to override any default value triggers etc.) it is suggested to name it z_[table_name]_[suffix].

**Concept:**

1. Not all triggers should be inherited;
2. There is no reason to create two or more triggers with different names, but calling the same trigger function on the same condition (`tgtype`);
3. There is no need to use `when` clause in triggers since all the checks can be done inside the trigger function (just for simplicity);
_4. There are situations when trigger function code might need to be adjusted (for example my case with `update` might use this feature); [not impletemented]_
_5. Triggers may change, thus a function is needed that can remove triggers and/or functions; [not impletemented]_
_6. New child-tables can be added and triggers should be applied to them automatically; [not implemented]_

As a result:

There's a table `referencing part_config(parent_name)` that keeps list of triggers for each parent table. It also stores the trigger type (`consistent` - same trigger function for all tables, `custom` - trigger function should be created for each table). For custom triggers an additional field is used that stores function source and name prefix. The point of custom functions is to adjust trigger code for each table (in my case for performance reasons).

Trigger creation is done in a stored procedure that uses a for cycle that checks if the trigger with same tgtype and function exists on each table of parittion.

Some changes need to be done to `run_maintenance` function to reflect the need to create triggers for newly created child-tables

User should specify the triggers that need to be inherited by adding a corresponding row to the `part_trigger` table.

Here's the code:

http://pastebin.com/w3jzQBDu

Keep in mind: for test case you need to create extension, then apply the patch and then proceed.

keithf4:
Looks very interesting so far! Just did a quick overview so far and didn't really dig into exactly how it all works, but just some notes for now.

- Please separate out your timezone changes from this. That is a different issue and any changes related to that should go into their own change set.

[...]

PR #43, opened 2014-12-23
Trigger Inheritance Implementation

jaggel:
I used the same approach as inherit fks so I added a column to
part_config (with default false though), implemented the
apply_triggers.sql and modified accordingly the create_parent and
create_time_partition to integrate the inherit_tg option.

Just trying to provide a solution / implementation for issue #42


keithf4:
Closing this pull request since it cannot be implemented. I will leave issue https://github.com/keithf4/pg_partman/issues/42 open until I figure something out for triggers, though.

PR #50, opened 2015-03-17
Added double quotes to to make privilege grantees delimited identifier

keithlyon:
Previous to this, if any compound role names (e.g. first.last) had been defined in the cluster, create_parent() would fail with ERROR:  syntax error at or near .


keithf4:
I have plans to do a whole rework of pg_partman to support reserved words & special characters in all object names, not just for role names. It's a lot of work, so it will take a while. Thanks, though!

PR #73, opened 2015-11-11
Fixed pattern for parent matching in reapply_indexes.py

edrep:
Fixed indexes being applied to the parent table instead of the children in reapply_indexes.py

### DIFF
--- a/bin/reapply_indexes.py
+++ b/bin/reapply_indexes.py
@@ -46,7 +46,7 @@ def create_index(conn, partman_schema, child_schemaname, child_tablename, child_
     parent_schemaname = result[0]
     parent_tablename = result[1]
     cur.close()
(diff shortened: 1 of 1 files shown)


keithf4:
Can you give me an example schema where you're seeing this happen? I haven't run across it applying any indexes to the parent and just want to understand better what may be causing the problem.

mariusperde:
Hi,

The regular expression didn't always match the parent statement because the last quote was not optional.

This is an example of a statement that wasn't be matched by the original expression (table *test* from the **public** schema):
```
CREATE INDEX test_text_idx ON test USING btree (text)
```

Failing to match the statement results in failing to replace " ON *table_name* " with " ON *child_table_name* " and the indexes end up on the parent table.

keithf4:
Ahh! I see. My testing didn't account for when the schema wasn't explicitly stated. Thanks for fixing the bug! I'll include it in my next release.

keithf4:
Pushed out an update with just fixes to this script. Let me know if this is working ok for you now.
Thanks!

edrep:
Yes, it's working as expected. Thank you.

PR #84, opened 2016-01-20
fix: do not work when partition table has no data

wonder365:
If the system halt for a few days, there is no data in partition table， run_maintenance do not create any new partitions.


keithf4:
Once you start inserting data again, run_maintenance should start making new partitions again. The data may go to the parent, but running partition_data_time() should fix that and get things running back to normal.

Seems different people want it to work different ways, so I'm going to be making it optional whether partitions continue being created when no data exists. No ETA on when I'll be getting that in, but appreciate the tip for where I can add the optional code in!

keithf4:
Version 2.3.0 has been released. There is now an option in the part_config table ( `infinite_time_partitions`) that will allow time-based partitions to continue being created infinitely, even with no new data. By default it is set to false, so just update the config table for the partition sets you want this to happen for.

Also, your push request didn't quite fix the issue properly. So if you implemented that in your own copy, I'd recommend looking how I did it or just update to 2.3.0.

PR #88, opened 2016-02-10
Fix conflict with intarray extension

lucasmo:
Without this fix, applying and dropping constraints does not work if the intarray extension is in the search path:
```
ERROR:  operator is not unique: smallint[] <@ smallint[]
LINE xx:         AND ARRAY[a.attnum] <@ con.conkey
```


keithf4:
Incorporated this into next version. Thanks for the fix!

keithf4:
This fix has been included in v2.3.0. Thanks!

PR #90, opened 2016-02-11
Use bigint instead of int for partition id in apply_constraints

lucasmo:
Fixes:
```
ERROR:  value "5499000000" is out of range for type integer
CONTEXT: PL/pgSQL function partman.apply_constraints(text,text,boolean,bigint,boolean) line 108 at SQL statement
```
I did a quick audit to see if this was an issue elsewhere and it doesn't appear to be.


keithf4:
Fixed in next release. Thanks again for all the bug fixes!

keithf4:
This fix has been included in v2.3.0. Thanks!

PR #99, opened 2016-03-10
Start partition number calculation fix

### LINKED ISSUE #98: Start partition number calculation
Hi!

Trying to run ```create_parent``` on existing table with lots of data I [found](https://github.com/keithf4/pg_partman/blob/master/sql/functions/create_parent.sql#L390-L395) it trys to run this: 
```sql
SELECT COALESCE('0', max('id')::bigint, 0) FROM "Schema"."Table" LIMIT 1
```
where ```'0'``` is value ```p_start_partition``` parameter. In my case this query executes in about 1.5-2 minutes. In combination with [access exclusive lock](https://github.com/keithf4/pg_partman/blob/master/sql/functions/create_parent.sql#L94) made before it creates some troubles. Is it really need to run this query when non-null ```p_start_partition``` is passed? If yes, maybe it's posstble to run this before locking table?

red-led:
Fixes #98

As I see this can't be moved before table lock, because ```v_top_parent_schema``` and ```v_top_parent_table``` are used, which should be determined after locking I believe. At least we can use ```p_start_partition``` without executing full-scan if it is set manually.

P.S. Sorry, I can't run test-case on my machine right now.

### DIFF
--- a/sql/functions/create_parent.sql
+++ b/sql/functions/create_parent.sql
@@ -387,12 +387,15 @@ IF p_type = 'id' THEN
 
     -- If custom start partition is set, use that.
     -- If custom start is not set and there is already data, start partitioning with the highest current value and ensure it's grabbed from highest top parent table
-    v_sql := format('SELECT COALESCE(%L, max(%I)::bigint, 0) FROM %I.%I LIMIT 1'
-                , p_start_partition::bigint
-                , p_control
-                , v_top_parent_schema
-                , v_top_parent_table);
-    EXECUTE v_sql INTO v_max;
+    IF p_start_partition IS NOT NULL THEN
+        v_max := p_start_partition::bigint;
+    ELSE
+        v_sql := format('SELECT COALESCE(max(%I)::bigint, 0) FROM %I.%I LIMIT 1'
+                    , p_control
+                    , v_top_parent_schema
+                    , v_top_parent_table);
+        EXECUTE v_sql INTO v_max;
+    END IF;
     v_starting_partition_id := v_max - (v_max % v_id_interval);
     FOR i IN 0..p_premake LOOP
         -- Only make previous partitions if ID value is less than the starting value and positive (and custom start partition wasn't set)


keithf4:
No worries on the test case. I don't merge code directly in through github unless it's made as a complete version upgrade. I'll check it out and add it in with the next patch set I'm working on. Thanks!

keithf4:
Also, concerning your documentation update, it should not be doing a full scan on the table if you have an index on the control column in place. If you don't have that index, I would highly recommend doing so otherwise you'll have major performance issues even after the partition set is created.

red-led:
In case where I have found this there was primary key on control column. Have no any other idea why it was so slow. To be absolutely sure I'll investigate into this with ```explain``` tomorrow.

red-led:
Ok. It's a little bit confusing me, but seems using constant in ```COALESCE``` is a problem.
Original query is going to scan whole table (runs in more then 1 mitute):
```SQL
EXPLAIN SELECT COALESCE('0', max(id)::bigint, 0) FROM "Schema"."Table" LIMIT 1
```
```
Limit  (cost=5339868.26..5339868.27 rows=1 width=0)
  ->  Aggregate  (cost=5339868.26..5339868.27 rows=1 width=0)
        ->  Seq Scan on "Table"  (cost=0.00..5339868.26 rows=102659026 width=0)
```

But if we just remove leading ```'0'``` in ```COALESCE``` magic happens:
```SQL
EXPLAIN SELECT COALESCE(max(id)::bigint, 0) FROM "Schema"."Table" LIMIT 1
```
```
Limit  (cost=0.85..0.86 rows=1 width=0)
  InitPlan 1 (returns $0)
    ->  Limit  (cost=0.57..0.85 rows=1 width=4)
          ->  Index Only Scan Backward using "Table_pkey" on "Table"  cost=0.57..28930418.86 rows=102659550 width=4)
                Index Cond: (id IS NOT NULL)
  ->  Result  (cost=0.00..0.01 rows=1 width=0)
```
Now query uses index and runs in less then 1msec!

This PR makes query look like second variant, so you're right, I should revert documentation update.

keithf4:
Hmm... looks like format() is being stupid and quoting that first parameter as a text value instead of leaving it bare integer. What plan does this give you?

```
EXPLAIN SELECT COALESCE('0'::bigint, max(id)::bigint, 0) FROM "Schema"."Table" LIMIT 1
```

keithf4:
Or if you can add that cast into the function itself here and try it, that would help me as well
https://github.com/keithf4/pg_partman/blob/master/sql/functions/create_parent.sql#L390

`v_sql := format('SELECT COALESCE(%L::bigint, max(%I)::bigint, 0) FROM %I.%I LIMIT 1'`

red-led:
@keithf4: I've tested it with ```'0'```, ```'0':bigint``` and ```0``` variants -- all have same plan.

keithf4:
They all do a seq scan if there's a third parameter at the beginning there? That doesn't make any sense if that value is actually an integer. Your first seq scan example made sense because it was assuming it was a text value and it's got an index on an integer value.

That first parameter is there as a check in case the p_start_partition argument is set, which is the whole reason for this issue.

keithf4:
Guess I'll just redo the logic there. I hadn't been running into any issues, even on really large tables, with doing `SELECT 0 FROM sometable LIMIT 1;`. Even though the plan says it's doing a seq scan, it seemed to do the smart thing and cut short actually scanning the whole table.

keithf4:
I've included your fix for the p_start_partition check in `create_parent()` for the next patch. Thanks!

red-led:
Yes, 
```SQL
EXPLAIN SELECT COALESCE(0, max(id)::bigint) FROM "Schema"."Table" LIMIT 1
```
has exactly same plan with all first argument versions. (LOL, ideally it should be instant, there is no matter whats in table).

This query runs in less than 1msec:
```SQL
SELECT 0 FROM  "Schema"."Table" LIMIT 1;
```
Plan:
```
Limit  (cost=0.00..0.05 rows=1 width=0)
  ->  Seq Scan on "Transactions"  (cost=0.00..5351660.28 rows=102885728 width=0)
```

I guess it may depend on some server customization like ```effective_io_concurrency``` or Planner Cost Constants.

P.S. ```PostgreSQL 9.5.1 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-4), 64-bit```

keithf4:
This has been incorporated into version 2.3.2. Closing Pull Request

PR #108, opened 2016-04-26
Fix security definer security issue

alvar-freude:
The following changes fixes a security issue with all functions using `SECURITY DEFINER` and adds a test for this case.

While running my test suite for an other project, I recognized that all functions with `SECURITY DEFINER` are created in a way, an attacker can misuse it.

For details see [Writing SECURITY DEFINER Functions Safely](http://www.postgresql.org/docs/current/static/sql-createfunction.html#SQL-CREATEFUNCTION-SECURITY) in the PostgreSQL documentation.

This pull request contains all changes and a test in `test/test-security--security-definer.sql` for this.

The fix is, that all `SECURITY DEFINER` functions need an explicit secure search path with pg_temp at the end. I added this in all functions.

Beside this, I include a very simple init&start script to start a test instance from scratch and a fix for `.gitignore`, which now includes the compiled SQL.

All tests are OK (tested with PostgreSQL 9.5), beside some tests which fail in the same way without my changes:

```
test/test-id-id-id-subpart.sql            (Wstat: 768 Tests: 1508 Failed: 0)
  Non-zero exit status: 3
  Parse errors: Bad plan.  You planned 3096 tests but ran 1508.
test/test-time-weekly-daily-subpart.sql   (Wstat: 0 Tests: 278 Failed: 4)
  Failed tests:  169-171, 173
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ ignore/*
 .deps/*
 *.o
 *.so
+sql/pg_partman--*.sql
--- a/sql/functions/apply_cluster.sql
+++ b/sql/functions/apply_cluster.sql
@@ -2,8 +2,8 @@
 * Function to apply cluster from parent to child table
 * Adapted from code fork by https://github.com/dturon/pg_partman
 */
-CREATE FUNCTION apply_cluster(p_parent_schema text, p_parent_tablename text, p_child_schema text, p_child_tablename text) RETURNS void
-    LANGUAGE plpgsql SECURITY DEFINER
+CREATE FUNCTION @extschema@.apply_cluster(p_parent_schema text, p_parent_tablename text, p_child_schema text, p_child_tablename text) RETURNS void
+    LANGUAGE plpgsql SECURITY DEFINER SET search_path = @extschema@, pg_temp
 AS $$
 DECLARE
     v_parent_indexdef   text;
--- a/sql/functions/apply_privileges.sql
+++ b/sql/functions/apply_privileges.sql
@@ -2,7 +2,7 @@
  * Apply privileges that exist on a given parent to the given child table
  */
 CREATE FUNCTION apply_privileges(p_parent_schema text, p_parent_tablename text, p_child_schema text, p_child_tablename text, p_job_id bigint DEFAULT NULL) RETURNS void
-    LANGUAGE plpgsql SECURITY DEFINER
+    LANGUAGE plpgsql SECURITY DEFINER SET search_path = @extschema@, pg_temp
     AS $$
 DECLARE
 
--- a/sql/functions/check_name_length.sql
+++ b/sql/functions/check_name_length.sql
@@ -3,8 +3,8 @@
  * Also appends given suffix and schema if given and truncates the name so that the entire suffix will fit.
  * Returns original name with schema given if it doesn't require truncation
  */
-CREATE FUNCTION check_name_length (p_object_name text, p_suffix text DEFAULT NULL, p_table_partition boolean DEFAULT FALSE) RETURNS text
-    LANGUAGE plpgsql SECURITY DEFINER
+CREATE FUNCTION @extschema@.check_name_length (p_object_name text, p_suffix text DEFAULT NULL, p_table_partition boolean DEFAULT FALSE) RETURNS text
+    LANGUAGE plpgsql SECURITY DEFINER SET search_path = @extschema@, pg_temp
     AS $$
 DECLARE
     v_new_length    int;
--- a/sql/functions/check_parent.sql
+++ b/sql/functions/check_parent.sql
@@ -2,7 +2,7 @@
  * Function to monitor for data getting inserted into parent tables managed by extension
  */
 CREATE FUNCTION check_parent() RETURNS SETOF @extschema@.check_parent_table
-    LANGUAGE plpgsql STABLE SECURITY DEFINER
+    LANGUAGE plpgsql STABLE SECURITY DEFINER SET search_path = @extschema@, pg_temp
     AS $$
 DECLARE
 
--- a/sql/functions/check_subpart_sameconfig.sql
+++ b/sql/functions/check_subpart_sameconfig.sql
@@ -21,7 +21,7 @@ RETURNS TABLE (sub_partition_type text
         , sub_optimize_constraint int
         , sub_infinite_time_partitions boolean
         , sub_jobmon boolean)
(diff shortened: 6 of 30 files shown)


keithf4:
Thanks! Completely missed that security tip in the docs. Not sure I'll include the initialization of test database at this time, but it's an interesting idea.

alvar-freude:
The test database is only a simple hack, which I used for this fix.
Usually I use my Perl Module `Test::PostgreSQL::Starter` for starting, initializing and stopping one or more instances, but this is not yet released on CPAN, so this was my fastest way to re-start everything ;-)

alvar-freude:
I reviewed my start/stop hack and my conclusion is, if you add them, then only in a folder "ugly-helper-hacks" or something similar … ;-)

keithf4:
Just realized setting the search path in this manner will break if pg_jobmon is installed. If you look at some function, I actually do alter the search path already (and don't even include pg_temp at all, which I believe does the same thing for the security issue in the end).

https://github.com/keithf4/pg_partman/blob/master/sql/functions/create_partition_id.sql#L59

So, better bet would be to change it to always do it in this manner for all the SEC DEF functions no matter if pg_jobmon is installed or not.

alvar-freude:
Ah, with pg_jobmon I haven't tested it. But setting the search_path should not  occur manually within each function, because it is slower. Maybe change the functions who need jobmon in search_path to change it at call time via ALTER FUNCTION?

keithf4:
I'd be curious how much slower that really is. It's doing a minimal catalog lookup for the current value then just setting it via a different call. Also, I did it this way because if someone is manually calling these functions, doing the path reset like this resets it for the whole session, not just for the run of the function. This was breaking things for myself while testing.

How would you change it at call time via ALTER FUNCTION?

keithf4:
So I've confirmed that the method of doing `SET search_path = @extschema@, pg_temp` does not work when pg_jobmon is installed. And I'd rather not redo things that make my pg_jobmon calls EXECUTE statements with a dynamic schema variable. And I have to do the search_path dynamically because I don't know what schema someone may have installed pg_jobmon in.

So for functions that don't use pg_jobmon, I'll be able to do this. But for those that do, I'll have to use my method of preserving the old search_path and resetting it. If you have a better way to do this, I'd look forward to hearing it.

alvar-freude:
hmm, OK – but what about call a `SET search_path = ...` via EXECUTE with a dynamic path? Or is it possible to change the search path after installation?

keithf4:
To change the search path within a function dynamically, you do what I've already been doing now using `set_config()`. Unfortunately, that changes the path either for the entire transaction (last parameter = true) or changes it for the entire session (last parameter = false). Either way, this can break things when you include pg_partman function calls in other transactions or sessions (my pgtap tests for example). This means you have to save the current search path and reset it at the end of the function run.

The method of doing SET search_path in the function definition is much better since it only sets the search_path for the length of the function run itself. Unfortunately, this cannot be made dynamic. So, I'll change things to do a mix of the two.

alvar-freude:
Is it not possible to call SET within a PL/pgsql EXECUTE?

alvar-freude:
See here: http://www.postgresql.org/docs/9.5/static/plpgsql-statements.html#PLPGSQL-STATEMENTS-EXECUTING-DYN

keithf4:
When I said "doing SET search_path in the function definition" I meant when you defined the search path after the SECURITY DEFINER declaration. Doing that seems to set the search_path for just the run of the function. Doing a `SET` call with `EXECUTE` within the function itself seems to reset the path for the transaction and/or session (haven't been able to determine which). That means it has the same issue I was already handling with `set_config()` 

Doing the `EXECUTE` statement with either `SET` or `set_config()` seems arbitrary. To me `set_config()` seems more clear what I'm doing. 

If you can get the SET method working without having to worry about resetting the search path back to the original value, and have it pass the pgtap tests with pg_jobmon installed, I'd be glad to try and include that method.

keithf4:
This is the new method I will be using that seems to work for me and allows the jobmon schema to be dynamic
```
DECLARE
[...]
v_new_search_path             text := '@extschema@,pg_temp';
v_old_search_path               text;
[...]
BEGIN
[...]

SELECT current_setting('search_path') INTO v_old_search_path;

IF v_jobmon THEN
    SELECT nspname INTO v_jobmon_schema FROM pg_catalog.pg_namespace n, pg_catalog.pg_extension e WHERE e.extname = 'pg_jobmon' AND e.extnamespace = n.oid;
    IF v_jobmon_schema IS NOT NULL THEN
        v_new_search_path := '@extschema@,'||v_jobmon_schema||',pg_temp';
    END IF;
END IF;

[...]

EXECUTE format('SELECT set_config(%L, %L, %L)', 'search_path', v_old_search_path, 'true');

**Comment by alvar-freude** [comment]
Sorry, I wrote misunderstanding stuff ;-)
I meant to write `ALTER FUNCTION ... SET search_path = ...`. 

So, the following is a temporary fix for me in my installation code, which takes an unmodified pg_partman and changes the pg_temp settings (but without any pg_jobmon search_path!):

```SQL
--
-- TEMP Security!
-- Change search_path for all SECURITY DEFINER functions of pg_partman
-- 

DO LANGUAGE plpgsql 
  $code$
    DECLARE r   RECORD;
    BEGIN
  
    FOR r IN 
          SELECT p.oid AS p_oid, 
                 nspname || '.' || proname AS name
            FROM pg_proc p
       LEFT JOIN pg_namespace n ON pronamespace = n.oid
           WHERE prosecdef
             AND nspname = 'partman'
    LOOP
      EXECUTE format('ALTER FUNCTION %s(%s) SET search_path = partman, pg_temp', r.name, pg_get_function_identity_arguments(r.p_oid));
    END LOOP;

    END;
  $code$;
```

keithf4:
Yeah unfortunately that's not going to work with pg_jobmon being able to be installed/uninstalled at any time. Also doing some more testing, all my `set_config()` calls above actually have to have the final parameter set to `false` otherwise it breaks other functions that may get called within the same transaction/session as pg_partman. 

So, I think I have a working fix for this in place. Just have to get some other issues sorted out.

alvar-freude:
Ah, ok, "being able to be installed/uninstalled at any time" is the part I was missing ;-)

keithf4:
I've got the better handling of SECURITY DEFINER added in v2.3.4. Thanks again for pointing this out!

alvar-freude:
Thanks for the fix and thanks for pg_partman! :)

PR #114, opened 2016-05-11
Fix check_parent for large databases

Crack:
When partman manages lots of tables it can fail with "out of shared memory" error because each `SELECT count(1)...` query creates an AccessShareLock and PostgreSQL keeps them until the end of session. In my case it breaks somewhere around 1550th table.

I made a workaround that uses BEGIN..END block with exceptions to simulate a SAVEPOINT...ROLLBACK sequence that allows PostgreSQL to release these locks.

A second change is an option to stop counting after first record in parent table is found. When I am looking for tables to fix with partition_data.py it doesn't matter how many records are there - it's only important to find tables that need attention.

I made a test with 1000 empty, partitioned tables - both versions (before and after my changes) run within 1.2s, so there's no easily measurable overhead due to throwing exceptions and doing rollbacks after each parent is checked.

### DIFF
--- a/sql/functions/check_parent.sql
+++ b/sql/functions/check_parent.sql
@@ -1,7 +1,7 @@
 /*
  * Function to monitor for data getting inserted into parent tables managed by extension
  */
-CREATE FUNCTION check_parent() RETURNS SETOF @extschema@.check_parent_table
+CREATE FUNCTION check_parent(p_exact_count boolean DEFAULT true) RETURNS SETOF @extschema@.check_parent_table
     LANGUAGE plpgsql STABLE SECURITY DEFINER
     AS $$
 DECLARE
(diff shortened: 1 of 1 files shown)


keithf4:
Very interesting stuff. Thanks for the patch!

For the limit query on the parent table, I changed it to do a `SELECT 1...` instead of `SELECT *` in the subquery which is a slight improvement since it then doesn't even have to do any further table lookups for any data.

For that EXCEPTION trick I just want to test it out a bit more and get some second opinions. It's a neat trick but just want to make sure it's doing what the intent is here. I'm assuming you've tested and got this working where you were getting the shared memory errors?

Crack:
`SELECT 1...` - IMHO ideally it should be `*` all the way, to tell PostgreSQL that we don't care about data at all, and we need only row counts - and in the future to allow for index-only scans which are currently unsupported. But I don't know whether `*` is safe - query plans are too abstract to be of any help here (I really miss some equivalent of Oracle's SQL trace'ing).

`EXCEPTION` - yes, I tested it on my data. Verified that in open transaction it actually clears locks, and that it allows me to run `check_parent` on my database (as opposed to failing on lock limit). Test environment used PostgreSQL 9.4 x64 on Windows. I also verified clearing of locks on my production database (9.4 on Debian).

keithf4:
New query:
```v_sql := format('SELECT count(1) AS n FROM (SELECT 1 FROM ONLY %I.%I LIMIT 1) x', v_schemaname, v_tablename);```

The thinking for doing this with `SELECT 1` instead of `SELECT *` is that with *, even if it's only one row, it still has to go to retrieve an actual tuple. If one of the columns is text, that may mean a TOAST table lookup as well. So, just trying to make this as performant as possible, which is the goal, this seemed better after some discussion with coworkers.

As far as `count(1)` vs `count(*)` I believe the two are equivalent internally as of 9.4.

keithf4:
I've added the p_exact_count option to `check_parent()` with v2.3.4. Haven't had time to test the other fix you submitted yet and just wanted to get this patch out first. Thanks!

keithf4:
Sorry I hadn't gotten back to you on that other fix with the ROLLBACK trick. Don't think I'm going to incorporate that at this time. The option to only get a single row back if it exists should hopefully help enough with performance issues when needing to check for rows in the parent table. That trick still seems a little hacky to me right now and not sure it's something to rely on going forward.

Thanks!

PR #120, opened 2016-06-10
"I wish u Happy debugging" bug fixed

madlynx:
create_partition_time() creates any time-based partition with parameters of first time-custom partition in system.


dturon:
This bug happend me and my friends many times ... sometimes just append AND something... but on previous line was WHERE a OR b :)

keithf4:
Thanks for the fix!

keithf4:
This has been fixed in v2.4.0. Thanks!

Issue #105, opened 2016-04-04
PGSQL 9.5 UPSERT support on partitioned tables.


MikaelUlvesjo:
In pgsql a new function is added to INSERT called "UPSERT" in the form of "ON CONFLICT".

I would like to use that feature on a partitioned table but since the trigger is not passing the ON CONFLICT part of the insert query it is ignored.

e.g:
When running the following queries on a table that is not partitioned both run fine and we will have 3 rows in the table, but if the table is partitioned the second query will fail and we will only have 2 rows in the table.

INSERT INTO part_table (id,val) VALUES 
(1,'test1'), (2,'test2')  ON CONFLICT (id) DO NOTHING

INSERT INTO part_table (id,val) VALUES 
(2,'test2'), (3,'test3')  ON CONFLICT (id) DO NOTHING

When running the following queries on a table that is not partitioned both run fine and we will have 3 rows in the table where row with id 2 will have the val = "test22", but if the table is partitioned the second query will fail and we will only have 2 rows in the table id=3 is missing and row with id = 2 will have val = "test2".

INSERT INTO part_table (id,val) VALUES 
(1,'test1'), (2,'test2')  ON CONFLICT (id) DO UPDATE SET val=EXCLUDED.val

INSERT INTO part_table (id,val) VALUES 
(2,'test22'), (3,'test3')  ON CONFLICT (id) DO UPDATE SET val=EXCLUDED.val

keithf4:
I'll see if I can add this in. Haven't done much testing against 9.5 features yet.
Keep in mind that any updates that would move a column from one child table to another (due to partitioning constraints) are not supported by pg_partman

MikaelUlvesjo:
I'm my case it would be enough to define it one time and the have it used in all insert queries.
I have a data source that will give me duplicate rows and if I could define "ON CONFLICT (id) DO NOTHING" once for the partitioning of a table it would solve my problem.
I have tried to manually update the trigger function to include the ON CONFLICT statement and it does work until a new version of the trigger is generated.

keithf4:
The issue I'm seeing now with this is how to managed the CONFLICT resolution. I have no way to predict how someone would want to resolve the issue and the possible resolutions are quite large. In your case DO NOTHING works fine. But that may not be what someone else needs to have done. Also there has to be a way to define which columns the CONFLICT should be handled for. There could be other unique constraints besides just a primary key.

I don't think there's going to be a quick resolution to this for right now unfortunately.

[...]

PR #122, opened 2016-06-16
added limited upsert support

MikaelUlvesjo:
A patch for "PGSQL 9.5 UPSERT support on partitioned tables. #105"

### DIFF
--- a/doc/pg_partman.md
+++ b/doc/pg_partman.md
@@ -78,7 +78,7 @@ As a note for people that were not aware, you can name arguments in function cal
 
 ### Creation Functions
 
-*`create_parent(p_parent_table text, p_control text, p_type text, p_interval text, p_constraint_cols text[] DEFAULT NULL, p_premake int DEFAULT 4, p_use_run_maintenance boolean DEFAULT NULL, p_start_partition text DEFAULT NULL, p_inherit_fk boolean DEFAULT true, p_epoch boolean DEFAULT false, p_jobmon boolean DEFAULT true, p_debug boolean DEFAULT false) RETURNS boolean`*
+*`create_parent(p_parent_table text, p_control text, p_type text, p_interval text, p_constraint_cols text[] DEFAULT NULL, p_premake int DEFAULT 4, p_use_run_maintenance boolean DEFAULT NULL, p_start_partition text DEFAULT NULL, p_inherit_fk boolean DEFAULT true, p_epoch boolean DEFAULT false, p_jobmon boolean DEFAULT true, p_debug boolean DEFAULT false, p_upsert text DEFAULT '') RETURNS boolean`*
 
  * Main function to create a partition set with one parent table and inherited children. Parent table must already exist. Please apply all defaults, indexes, constraints, privileges & ownership to parent table so they will propagate to children.
  * An ACCESS EXCLUSIVE lock is taken on the parent table during the running of this function. No data is moved when running this function, so lock should be brief.
@@ -120,7 +120,11 @@ As a note for people that were not aware, you can name arguments in function cal
  * `p_epoch` - tells `pg_partman` that the control column is an integer type, but actually represents and epoch time value. All triggers, constraints & table names will be time-based. Be sure you create a functional, time-based index on the control column (to_timestamp(controlcolumn)) so this works efficiently.
  * `p_jobmon` - allow `pg_partman` to use the `pg_jobmon` extension to monitor that partitioning is working correctly. Defaults to TRUE.
  * `p_debug` - turns on additional debugging information.
-
+ * `p_upsert` - adds upsert to insert queries in triggers, to allow handeling of conflicts  Defaults to ''
+    + to ignore conflicting rows on a table with primary key id set p_upsert to `'ON CONFLICT (id) DO NOTHING'`
+	+ to update a conflicting row on a table with columns id(pk), val set p_upsert to `'ON CONFLICT (id) DO UPDATE SET val=EXCLUDED.val'`
+	+ Requires postgresql 9.5
+	+ See https://www.postgresql.org/docs/9.5/static/sql-insert.html 
 
 *`create_sub_parent(p_top_parent text, p_control text, p_type text, p_interval text, p_constraint_cols text[] DEFAULT NULL, p_premake int DEFAULT 4, p_start_partition text DEFAULT NULL, p_inherit_fk boolean DEFAULT true, p_epoch boolean DEFAULT false, p_jobmon boolean DEFAULT true, p_debug boolean DEFAULT false) RETURNS boolean;`*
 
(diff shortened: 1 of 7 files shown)


keithf4:
Very cool! Thanks for all the work. Been pretty busy lately, so won't have this ready to merge into the next version I'm already working on. Definitely want to look this over and get it in if possible in the near future, though.

bk1te:
Hi any news on this issue?

keithf4:
I do plan on integrating once I have the time to review it. However bug fixes are higher priority for me than new features, so working on those first.

keithf4:
So, implemented your changes so far (which some modification) to the "upsert" branch. 

https://github.com/keithf4/pg_partman/tree/upsert

You can check it out to see my changes. One big thing I noticed was your examples were missing the DO UPDATE part of the ON CONFLICT clause. That was easy to fix. 

Bigger issue I've run into is with subpartitioning, which hadn't been handled either. You can see the issue in the `test-id-time-subpart-upsert-do-nothing.sql` test. I have it stop just after the error and leave things as they are in the database without rolling back. Because the date values are different, they go into different time sub-child tables, but allow duplication of the primary key value
```
keith@keith=# select * from partman_test.id_taptest_table_p10;
 col1 |  col2  |             col3              
------+--------+-------------------------------
   10 | «NULL» | 2016-07-07 18:17:10.858135-04
   11 | «NULL» | 2016-07-07 18:17:10.858135-04
   12 | «NULL» | 2016-07-07 18:17:10.858135-04
   13 | «NULL» | 2016-07-07 18:17:10.858135-04
   14 | «NULL» | 2016-07-07 18:17:10.858135-04
   15 | «NULL» | 2016-07-07 18:17:10.858135-04
   16 | «NULL» | 2016-07-07 18:17:10.858135-04
   17 | «NULL» | 2016-07-07 18:17:10.858135-04
   18 | «NULL» | 2016-07-07 18:17:10.858135-04
   19 | «NULL» | 2016-07-07 18:17:10.858135-04
   10 | «NULL» | 2016-07-08 18:17:08.299119-04
   11 | «NULL» | 2016-07-08 18:17:08.299119-04
   12 | «NULL» | 2016-07-08 18:17:08.299119-04
   13 | «NULL» | 2016-07-08 18:17:08.299119-04
   14 | «NULL» | 2016-07-08 18:17:08.299119-04
   15 | «NULL» | 2016-07-08 18:17:08.299119-04
   16 | «NULL» | 2016-07-08 18:17:08.299119-04
   17 | «NULL» | 2016-07-08 18:17:08.299119-04
   18 | «NULL» | 2016-07-08 18:17:08.299119-04
   19 | «NULL» | 2016-07-08 18:17:08.299119-04
(20 rows)

Time: 5.890 ms
keith@keith=# select * from partman_test.id_taptest_table_p10_p2016_07_07;
 col1 |  col2  |             col3              
------+--------+-------------------------------
   10 | «NULL» | 2016-07-07 18:17:10.858135-04
   11 | «NULL» | 2016-07-07 18:17:10.858135-04
   12 | «NULL» | 2016-07-07 18:17:10.858135-04
   13 | «NULL» | 2016-07-07 18:17:10.858135-04
   14 | «NULL» | 2016-07-07 18:17:10.858135-04
   15 | «NULL» | 2016-07-07 18:17:10.858135-04
   16 | «NULL» | 2016-07-07 18:17:10.858135-04
   17 | «NULL» | 2016-07-07 18:17:10.858135-04
   18 | «NULL» | 2016-07-07 18:17:10.858135-04
   19 | «NULL» | 2016-07-07 18:17:10.858135-04
(10 rows)

Time: 0.951 ms
keith@keith=# select * from partman_test.id_taptest_table_p10_p2016_07_08;
 col1 |  col2  |             col3              
------+--------+-------------------------------
   10 | «NULL» | 2016-07-08 18:17:08.299119-04
   11 | «NULL» | 2016-07-08 18:17:08.299119-04
   12 | «NULL» | 2016-07-08 18:17:08.299119-04
   13 | «NULL» | 2016-07-08 18:17:08.299119-04
   14 | «NULL» | 2016-07-08 18:17:08.299119-04
   15 | «NULL» | 2016-07-08 18:17:08.299119-04
   16 | «NULL» | 2016-07-08 18:17:08.299119-04
   17 | «NULL» | 2016-07-08 18:17:08.299119-04
   18 | «NULL» | 2016-07-08 18:17:08.299119-04
   19 | «NULL» | 2016-07-08 18:17:08.299119-04
(10 rows)

Time: 0.403 ms
```
Right now, I don't see any way around this since it's basically an issue with the core functionality of inheritance and the lack of global constraints. Only solution I can see right now is just not allowing the upsert option with subpartition sets. I've already limited other subpartition functionality, so I don't think this would be a huge deal. If you've got any ideas, I'd be glad to hear them.

keithf4:
Actually, just realized after I posted this that that's an issue in general even without the UPSERT. But I think allowing the UPSERT option on subpartitioning would make it more confusing since the expected behaviour would be to not allow that at all.

bk1te:
I think when developer uses specific features of database, like UPSERT, he is responsible of looking after constraints. Without upsert support in pg_partman there is no possibility to use it in specific applications. For now i am using PR in development...

keithf4:
I'm all for getting the feature in, my worry is just with subpartitioning. UPSERT is entirely reliant on constraint management working a certain way and it seems a bit of a footgun with multiple-level inheritance getting involved since there's a higher expectation of things working a certain way.

Thinking I'm just not going to allow it to work with sub-partitioning for now so I can get the feature in and maybe down the line a better solution will come around.

MikaelUlvesjo:
I was aware of the subpartition problem and that is why I added this with the header "limited upsert support". I agree with bk1te that if a developer uses functions like upsert he/she will have to know what they are doing, using the do update might be a problem with subpartition but do nothing function should work. BTW I'm sure that I wrote both do update and do nothing examples, I might have done that after the commit. I'm on vacation now and will not be able to check until the end of next week.

MikaelUlvesjo:
Did not mean to close...

keithf4:
You did write both examples. No worries there.

My example that failed with subpartitioning is actually a DO NOTHING example. It actually has nothing to do with the upsert feature at all. You can trigger the problem in the current version of pg_partman as well. My issue is the fact that someone going out of their way to use the upsert feature has a higher expectation of how constraints will be managed. Using upsert with multi-level inheritance is much more prone to issues since data will be getting written with the expectation that constraint violations will be handled in a specific way and, in all likelihood, they will not.

keithf4:
Just thinking about this more, it's still a problem in single-level partitioning as well. If a table is partitioned by time and someone inserts the same primary key value again later that would put it in a different child, and it will allow that go go through without throwing an error. 

So, this only works for any constraint violations (not just primary keys) within each individual child table itself.  Data duplication can easily run rampant, but that can generally be handled by using the monitoring function I provide. However many DO UPDATE clauses for affected rows simply wouldn't fire at all. I'm really wondering how useful it actually is if conflicts are not handled across the entire partition set properly.

Have you all thought this out and accept that UPSERT simply does not work as intended with inheritance? I don't run anything in production where UPSERT is needed in a partitioned set, but you do. Going to be some major caveats on using this feature if implemented.

MikaelUlvesjo:
I have tried it on a test environment but with real data stream, with id partitioning, both with do nothing and do update and in the do update case I did not change the id that is used in the partitioning only other data, and it worked perfectly, even with quite complicated do update queries where fields only where updated if the row in the db contained null and the conflicting row had a value. So I used it to do an update query and not a insert query when a conflict was detected, this way the application it self did not have to care if the new data was an update or an insert and this boosted the performance of the app quite a bit. The software I'm writing gets about 200 rows a second 24/7 so it is important that the performance and load on the system is not loading the system more then needed and that even when the table is locked for writing it will manage to insert all the data collected during that time fast so the queue don't build up over time.

keithf4:
Ok. I'll see about getting it in. Just have to word the documentation carefully.

Thanks for the input!

keithf4:
@MikaelUlvesjo @bk1te I've pushed another update out to the upsert branch

https://github.com/keithf4/pg_partman/tree/upsert

Can you please test this out and see if things work ok for you? Since the reasons it was failing with subpartitioning are the same shortcomings it has with normal partitioning, I didn't disable it there. Added some checks in to ensure the option is only usable in 9.5. Fixed some bugs encountered when running full pgtap test suite and updated the examples for more clarity.

Also would appreciate any feedback on the clarity of my explanations of the limitations upsert has with partitioning in the doc file.

Thanks!

MikaelUlvesjo:
I looked at you commit and can't find anything in the code that I have any comments on.

In the documentation on ON CONFLICT DO UPDATE we might want to add a text that says something like this.

When using ON CONFLICT DO UPDATE it is important to NOT change the value that the partitioning is based on as the decision on what partition the data will be in is already taken and if changed to a value that belongs to another partition that row will end up in the wrong partition.

keithf4:
Forgot about that. Thanks!

keithf4:
I've released version 2.5.0 that incorporates this feature. Let me know if things are working ok.

MikaelUlvesjo:
Latest release I can find is 2.4.1, did you forgot to do the actual release?

keithf4:
I've pushed the code up. I usually ask the person who's feature/bug I added to check it out before I tag the release since it don't want to change things in a tagged release without releasing a new version.

MikaelUlvesjo:
I have looked at the code any run it manually in my own test environment, but I can't test it on the test server without a real version that I can add to the build-scripts config.
You have my OK, i have done the testing I can.

keithf4:
Version 2.5.0 tagged. Thanks!

PR #123, opened 2016-06-16
Support INSERT Statement with RETURNING clause

zerojunks:
My ORM library requires returning clause so i made a small patch.
I applied to my production site and had no problems till now.
Thanks for your nice solution.

### DIFF
--- a/make_without_bgw.bat
+++ b/make_without_bgw.bat
@@ -1,0 +1,7 @@
+@echo Minimum version of PostgreSQL required is 9.4.0
+@mkdir temp
+bitsadmin.exe /transfer "Downloading BusyBox" https://frippery.org/files/busybox/busybox.exe "%~dp0temp\busybox.exe"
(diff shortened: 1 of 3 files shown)


keithf4:
If I understand correctly, this makes every insert to all partition sets always provide a RETURNING statement. Unfortunately, I don't think that's a great solution at this time since whether or not you need RETURNING values is usually on a case by case basis, not an all or nothing solution.

Could maybe see about making an option per partition set, but even then, you're still going with all or nothing for that partition set.

Right now there just aren't any good general solutions to this problem outside of core postgres having better options with triggers rewriting things. Thankfully your patch isn't too complicated and I can't foresee anything making it too difficult to maintain your own fork with this feature.

zerojunks:
I thought always providing returning statement is acceptable for the compatibility.
But it seems like this is not an essential, just optional feature.
l'll keep developing in my fork.

PR #125, opened 2016-06-23
Update of file partition_data_id.sql

thomasboussekey:
In the function `partition_data_id`, enlarge storage type for parameters:
* p_batch_count 
* p_batch_interval

From `INTEGER` To `Bigint`

### DIFF
--- a/sql/functions/partition_data_id.sql
+++ b/sql/functions/partition_data_id.sql
@@ -1,7 +1,7 @@
 /*
  * Populate the child table(s) of an id-based partition set with old data from the original parent
  */
-CREATE FUNCTION partition_data_id(p_parent_table text, p_batch_count int DEFAULT 1, p_batch_interval int DEFAULT NULL, p_lock_wait numeric DEFAULT 0, p_order text DEFAULT 'ASC') RETURNS bigint
+CREATE FUNCTION partition_data_id(p_parent_table text, p_batch_count bigint DEFAULT 1, p_batch_interval bigint DEFAULT NULL, p_lock_wait numeric DEFAULT 0, p_order text DEFAULT 'ASC') RETURNS bigint
     LANGUAGE plpgsql SECURITY DEFINER
     AS $$
 DECLARE


keithf4:
Can almost understand the batch_interval needing to to be that big, but are you really running over 2 billion batches? I kinda set that data type as a sanity check.

keithf4:
Just thinking about this more, not even sure allowing a bigint for the batch_interval is wise. Doing a commit with more than 2billion rows in a single transaction just sounds like a bad idea. If you can give me a working example where this is needed, I'll reconsider, but for now, I'm going to leave these data types as is just to hint people that anything larger is probably not a good idea.

thomasboussekey:
Hello,

I will clarify my partitioning scheme, it will help us!

# Context: Why do I need `BIGINT` id partitioning?

I'm working for a retail company that has **more than a thousand** of point of sales **POS**:
* `physical store` 
* `Internet e-commerce website`

In our `master data management system`, our point of sales are identified by 3 numbers:
* Type of entity (max 2 digits, this number is different for `physical store` and `Internet e-commerce website`)
* Number of entity (Today, max 4 digits, but we want to have the possibility to extend it to 5 digits)
* Sub-number of entity (Today, max 4 digits, but we want to have the possibility to extend it to 5 digits)

So, we defined a `human readable` identifier for partitioning agregating these 3 numbers, plus a final "0" for aligning the ID with the partition name.


This is the reason why we need to perform partitioning on a `BIGINT` data type.

# Partitioning scheme.

I want to reassure you, we don't want to manage billions of partitions.
Our production application will shard data on several different databases.
So, our table will have to manage `customer invoices` for hundreds of `POS`.

For technical purpose, and performance, we want to partition our tables according to the `POS` identifier.

So, we only need **hundreds** of partition per tables on specific value.
Our goal is not to try to manage 2 billions of partitions.

keithf4:
My apologies... I was getting the batch interval mixed up with how many rows are moved per batch. I'll see if I can get the data type for the batch_interval changed for the next version. However, I believe the count should still say an INT.

keithf4:
I've fixed the interval to be a bigint in v2.4.1. Let me know if you have any further issues!

PR #137, opened 2016-09-01
Added p_analyze parameter to create_function_id, create_parent, parti…

MikaelUlvesjo:
Added p_analyze parameter to create_function_id, create_parent, partition_data_id and partition_data_time to be able to control if the analyze should run when a new child table is created.

We have the problem that when a new sub partition is created a ANALYZE is run on the parent table and when this is performed the table is locked and the analyze will take very long time and not allowing us to use the table at all, when running the analyze manually the table is not locked.

Now I added parameter to all places that I could find where a call to create_partition_id or create_partition_time is potentially invoked to make sure that it is possible to disable the analyze no matter how the new sub_table is created.

### DIFF
--- a/doc/pg_partman.md
+++ b/doc/pg_partman.md
@@ -82,7 +82,7 @@ As a note for people that were not aware, you can name arguments in function cal
 
 ### Creation Functions
 
-*`create_parent(p_parent_table text, p_control text, p_type text, p_interval text, p_constraint_cols text[] DEFAULT NULL, p_premake int DEFAULT 4, p_use_run_maintenance boolean DEFAULT NULL, p_start_partition text DEFAULT NULL, p_inherit_fk boolean DEFAULT true, p_epoch boolean DEFAULT false, p_upsert text DEFAULT '', p_trigger_return_null boolean DEFAULT true, p_jobmon boolean DEFAULT true, p_debug boolean DEFAULT false)`*
+*`create_parent(p_parent_table text, p_control text, p_type text, p_interval text, p_constraint_cols text[] DEFAULT NULL, p_premake int DEFAULT 4, p_use_run_maintenance boolean DEFAULT NULL, p_start_partition text DEFAULT NULL, p_inherit_fk boolean DEFAULT true, p_epoch boolean DEFAULT false, p_upsert text DEFAULT '', p_trigger_return_null boolean DEFAULT true, p_jobmon boolean DEFAULT true, p_debug boolean DEFAULT false, p_analyze boolean DEFAULT true)`*
 
  * Main function to create a partition set with one parent table and inherited children. Parent table must already exist. Please apply all defaults, indexes, constraints, privileges & ownership to parent table so they will propagate to children.
  * An ACCESS EXCLUSIVE lock is taken on the parent table during the running of this function. No data is moved when running this function, so lock should be brief.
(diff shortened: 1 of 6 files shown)


keithf4:
There is already an option to the run_maintenance() function to control whether an analyze is run when new partitions are created. There's also a GUC for postgresql.conf when using the BGW. Are you still seeing the analyze triggered somewhere? The places you added the option don't seem relevant to normal maintenance.

I specifically removed the analyze from the create_parent() function a while ago, so it should never be doing that when a new partition set is created. The options on that function are already getting out of hand and I don't really see a need for it when future maintenance (unless specified otherwise) should run the analyze.

Not sure why you added the option to create_function*. That does not trigger an analyze that I can see. Did I miss something?

I could see adding that option to the partition_data* functions, and I'll see about doing that during my next update. But those are not called during normal maintenance.

MikaelUlvesjo:
Hi

I added the option to create_function_id as it will create a trigger that
will create sub tables that will run analyze if bgw is disabled.

Create_parent calls create_function_id and need the parameter as described
above.

Partition_data_id calls create_function and needs it as described above and
it calls create_partition_id that without the parameter will run analyze.

Partition_data_time do not need is I made a mistake with that one.

runt maintenance needs it as it calls create_function_id described above.

In our setup we need to run Partition_data_id as we seen that we get rows
in the parent table that needs to be moved to the correct partition.
We had problem when we did not use the bgw so we switched to use it but
with this patch we did not have to do that.

Den 1 sep. 2016 4:31 em skrev "Keith Fiske" <notifications@github.com>:

keithf4:
Ahh I see where it's doing that for create_function_id now. I'd missed that, thanks.

I'd just make create_parent() send false by default instead of adding a new parameter.

I've been debating removing that 50% table creation option for ID partitioning. With the BGW available now, it's not quite as critical since an external scheduler isn't needed. It seems to just be causing more and more edge cases I have to deal with. Not going to do that until I get a C version of run_maintenance completed, so will be a while. But thought I'd let you know and see your opinion.

MikaelUlvesjo:
Hi

With your suggested changes to create_function_id with default false and
adding parameter to partition_data_id I'm happy :-) it solves both issues I
have had.

Den 1 sep. 2016 8:22 em skrev "Keith Fiske" <notifications@github.com>:

keithf4:
Great. Not sure I'll have this fix out in the near future. I'm doing work on the C function now to see what I may be able to get done for a new release. If it drags on too long, I'll definitely do a patch release with these features. Kinda why I did the 2.6.0 release recently. C function was taking more work than I thought. :)

keithf4:
The below Gist contains the next patch update for pg_partman and includes this fix. If you've got time, please test it out in the environment where you encountered the performance issue and let me know how it goes. It's not the final version, so please don't install it on your prod system yet. I'll hopefully have the update out next week sometime then.

https://gist.github.com/keithf4/e86bafa682f086ad90b67e2240281f7d

MikaelUlvesjo:
Thanks, I will see if I can get time next week to look at it.

Den 23 sep 2016 20:30 skrev "Keith Fiske" <notifications@github.com>:

keithf4:
I've pushed up v2.6.1 with these changes.
Waiting on tagging a new version while I evaluate it further on current systems.

MikaelUlvesjo:
Ok, I have not had a chance to test it and probably won't be able to test
it the next two weeks. It's hard for me to test it before it is tagged as
our build path require a tagged version.

Den 30 sep. 2016 6:56 em skrev "Keith Fiske" <notifications@github.com>:

I've pushed up v2.6.1 with these changes.
Waiting on tagging a new version while I evaluate it further on current
systems.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/keithf4/pg_partman/pull/137#issuecomment-250796544>,
or mute
the thread

keithf4:
v2.6.1 tag has been released

PR #140, opened 2016-09-26
background worker fixes

saaros:
Fixes for two issues I spotted while reviewing the code.

### DIFF
--- a/src/pg_partman_bgw.c
+++ b/src/pg_partman_bgw.c
@@ -265,7 +265,14 @@ void pg_partman_bgw_main(Datum main_arg) {
                 worker.bgw_main = NULL;
                 sprintf(worker.bgw_library_name, "pg_partman_bgw");
                 sprintf(worker.bgw_function_name, "pg_partman_bgw_run_maint");
-                sprintf(worker.bgw_name, "pg_partman dynamic background worker (dbname=%s)", dbname);
+                rc = snprintf(worker.bgw_name, sizeof(worker.bgw_name),
+                              "pg_partman dynamic background worker (dbname=%s)", dbname);
+                if (rc >= sizeof(worker.bgw_name)) {
+                    /* dbname was truncated, add an ellipsis to denote it */
+                    const char truncated_mark[] = "...)";
+                    memcpy(worker.bgw_name + sizeof(worker.bgw_name) - sizeof(truncated_mark),
+                           truncated_mark, sizeof(truncated_mark));
+                }
                 worker.bgw_main_arg = Int32GetDatum(dbcounter);
                 worker.bgw_notify_pid = MyProcPid;
 
--- a/src/pg_partman_bgw.c
+++ b/src/pg_partman_bgw.c
@@ -400,7 +400,10 @@ void pg_partman_bgw_run_maint(Datum arg) {
             , dbname);
 
     resetStringInfo(&buf);
-    appendStringInfo(&buf, "SELECT n.nspname FROM pg_catalog.pg_extension e JOIN pg_catalog.pg_namespace n ON e.extnamespace = n.oid WHERE extname = 'pg_partman'");
+    appendStringInfo(&buf,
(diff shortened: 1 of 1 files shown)


keithf4:
Thanks very much! I'm working on a new patch release right now and will incorporate these two changes.

keithf4:
Curious on your reuse of the `rc` variable with snprintf. It's used earlier to store a return code value for the latch process. Don't think re-using it will hurt anything, but confused me for a bit there when I was checking if it would. Probably be better to use its own variable to avoid confusion.

saaros:
I thought the `rc` name was generic enough that it could be reused and that nothing cared about the rc value anymore at that point in the code, but I can update my commit to use a new variable if you prefer that.

keithf4:
No problem. Just wanted to make sure I wasn't missing something. I'll take care of using a new variable when I incorporate it.

keithf4:
So looking at the issue more, it's not stuck in an unintentional loop. It's just running the main background worker loop that starts the other dynamic ones really, really fast. Like the latch wait isn't working. I got that code from the worker_spi and that still seems to be working fine in 9.6.

keithf4:
Have you tried testing that last commit you did for quoting the schema name? I keep getting odd syntax errors when I run the pgtap tests when it calls run_maintenance()
```
2016-09-28 14:16:34 EDT [] [7278]: [9-1] user=,db=,e=42601 ERROR:  syntax error at or near "" at character 8
2016-09-28 14:16:34 EDT [] [7278]: [10-1] user=,db=,e=42601 QUERY:  SELECT partman~.run_maintenance(p_analyze := true, p_jobmon := true)
2016-09-28 14:16:34 EDT [] [4836]: [722-1] user=,db=,e=00000 LOG:  worker process: pg_partman dynamic background worker (dbname=keith) (PID 7278) exited with exit code 1
```
Tried doing it with just throwing quotes around it and get similar errors
```
    appendStringInfo(&buf, "SELECT '\"'|| n.nspname ||'\"' FROM pg_catalog.pg_extension e JOIN pg_catalog.pg_namespace n ON e.extnamespace = n.oid WHERE extname = 'pg_partman'");


2016-09-28 14:25:01 EDT [] [7640]: [9-1] user=,db=,e=42601 ERROR:  syntax error at or near "" at character 8
2016-09-28 14:25:01 EDT [] [7640]: [10-1] user=,db=,e=42601 QUERY:  SELECT "partman"~.run_maintenance(p_analyze := true, p_jobmon := true)
```

Only other solution I see that does work right now is double-quoting whenever the partman schemaname is used
```
    appendStringInfo(&buf, "SELECT \"%s\".run_maintenance(p_analyze := %s, p_jobmon := %s)", partman_schema, analyze, jobmon);
```

keithf4:
FYI, I've created a separate issue specifically for the 9.6 compatibility problems. Feel free to submit patches to this pull request, but if you have any fixes regarding that specifically, please post to that issue as well. Thanks!

https://github.com/keithf4/pg_partman/issues/141

keithf4:
Removed another extraneous `continue` from `pg_partman_bgw_main()`, but doesn't seem to have fixed the 9.6 issue yet
```
         } else { // pg_partman_bgw_dbname if null
             elog(DEBUG1, "pg_partman_bgw.dbname GUC is NULL. Nothing to do in main loop.");
         }
-        continue;
```

keithf4:
I've incorporated all the fixes you've posted so far into 2.6.1. As I said above, I was getting errors with your quote_ident() method of handling non-standard schema names, so I just did it the way I mentioned for now.
Currently waiting on tagging the release while I evaluate it on some systems further.

keithf4:
I believe I've fixed the issue with 9.6 in version 2.6.3. Closing this task for now. If you have any other patches for the background worker, feel free to create a new Issue or Pull Request at any time. Thanks!

ChristophBerg:
The regression failures I've seen in Debian are fixed, thanks!

PR #143, opened 2016-10-31
Parameterizing epoch logic

Trekoid:
What do you think of the following refactor that will eliminate the large duplicative code blocks for the epoch logic? It will significantly shrink the size of the time range functions and make the code easier to maintain. If this is acceptable, I will apply the same change to the rest of the `if v_epoch` conditionals.

Issue #142, opened 2016-10-28
Partitioning by tstzrange


Trekoid:
I'm on a project where range partitioning by the lower bound of a tstzrange column is needed. We plan on implementing by using pg_partman as a basis since lower(tstzrange) = timestampz. Is this a feature that would be desired such that we should contribute it back to the project or is it outside of the scope of the planned roadmap?

keithf4:
pg_partman is centered around partitioning predetermined intervals on a static value column. Adding an option just for lower(tstzrange) is really too narrow a use case for adding that kind of complexity to the trigger generation. If you need something a little more flexible, you could take a look at the range_partition extension I mention in the README.

If you'd like to use pg_partman, another option would be to have another column that is automatically populated with the lower bound of the tstzrange column value via a trigger. You could then partition based on that generated column since it's a static time value. But if the lower value of that range ever changes, that wouldn't work since that could possibly break the constraint and require moving the data to another child table. Something to keep in mind with whatever solution you go with.

Trekoid:
For my use case, lower(tstzrange) is static. We are doing something akin to session logging. The start time is fixed, the end time may extend for a period of time. After 24 hours all records are stale and do not need to be in the active table anymore, so we wanted to have a daily window of active data.

I have looked at range_partition, pg_slice, pg_pathman and pg_partman. The issue all these partition tools have is they work off of a column of type timestamp. In the broadest sense, I am looking to partition off of an expression that evaluates to type timestamp. I haven't found a tool yet that meets this need. I guess adoption of the native PostgreSQL data range types is still in the early upswing.

I considered having the extra column, but I disliked the idea of having duplicate data. There is also the option to break the tstzrange into two columns, an open and a close. However, I lose the easy query syntax—such as the `@>` operator—and most importantly I cannot setup an exclusion constraint to prevent overlapping sessions—`exclude using gist (id_col with =, tstzrange_col with &&)`.

I decided to start with pg_partman as a base as all of the time range partitioning logic is there. I would just need to tweak things a bit to include the lower() function call to extract the relevant timestamp out of the range data type. The trigger function would end up looking something like: 
`       IF TG_OP = 'INSERT' THEN 
            v_partition_timestamp := date_trunc('day', lower(NEW.col3));
            IF lower(NEW.col3) >= '2016-10-27 00:00:00-07' AND lower(NEW.col3) < '2016-10-28 00:00:00-07' THEN 
[...]

PR #149, opened 2016-11-22
Partition by tsrange or tstzrange with a static bound

Trekoid:
With the refactor from #143 in place, the ability to partition using the bounds of a tsrange or a tstrange as mentioned in #142, has become easy to add. We have started using the attached changes internally at Porch.com to partition by the lower bound of a tstzrange data type column.

Others may want to partition by a static upper bound, so a complete feature implementation would be to add a parameter that would allow that to be configured. The case statement would end up looking like the following with a parameter added. The default for p_upper_range would be false, so that it would be an optional argument, providing backwards compatibility.

    v_partition_expression := case
        when v_epoch then format('to_timestamp(NEW.%I)', v_control)
        when v_ranged_control and p_upper_range then format('upper(NEW.%I)', v_control)
        when v_ranged_control and not p_upper_range then format('lower(NEW.%I)', v_control)
        else format('NEW.%I', v_control)
    end;

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,9 @@
+2.6.2
+-- Refactor of time functions to improve maintainability of the code.
+   -- A new variable v_partition_expression is set with the appropriate substitution based upon p_epoch configuration
+   -- If-else blocks used to branch execution for epoch support were eliminated by using v_partition_expression in string substitutions
+
+
 2.6.1
 -- WARNING: The pg_partman background worker is currently incompatible with PostgreSQL 9.6. It runs fine in 9.4 and 9.5. I'm still working on trying to figure out what the issue is. If you must upgrade to 9.6 before this issue is fixed, you can disable the background worker by removing the entry for it in shared_preload_libraries and restarting postgres. Instead schedule the run_maintenance() plpgsql function to run via another scheduler method (cron).
 
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,7 +1,6 @@
 2.6.2
--- Refactor of time functions to improve maintainability of the code.
-   -- A new variable v_partition_expression is set with the appropriate substitution based upon p_epoch configuration
-   -- If-else blocks used to branch execution for epoch support were eliminated by using v_partition_expression in string substitutions
+-- Refactor of time functions in relation to epoch partitioning to improve maintainability of the code. Much thanks to Michael Rasmussen (Trekoid)! (Github Pull Request #143)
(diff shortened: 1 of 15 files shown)


Trekoid:
Having had a chance to step back, another approach would be to configure the partition expression based upon the data type of the control column. Instead of a boolean v_ranged_control variable, there would be a v_control_data_type variable. The data type of the control variable could be added to the configuration table when partitioning was initialized, so the select wouldn't have to be run in all of the functions.

    v_control_data_type    VARCHAR;
    
    SELECT data_type
    INTO v_control_data_type
    FROM information_schema.columns
    WHERE table_schema = v_parent_schema
    AND table_name = v_parent_tablename
    AND column_name = v_control;

    v_partition_expression := case
        -- integer and floating types that can be treated as epoch and converted using the to_timestamp function
        when v_control_data_type in ('smallint', 'integer', 'bigint'
                , 'decimal', 'numeric', 'real', 'double precision')
            then format('to_timestamp(NEW.%I)', v_control)
        when v_control_data_type in ('tsrange', 'tstzrange') and p_upper_range
            then format('upper(NEW.%I)', v_control)
        when v_control_data_type in ('tsrange', 'tstzrange') and not p_upper_range 
            then format('lower(NEW.%I)', v_control)
        when v_control_data_type in ('timestamp', 'timestamp with time zone', 'date')
            then format('NEW.%I', v_control)
        else
            raise error 'Unsupported data type detected'
    end;

keithf4:
Any chance you could do this as another pull request with all the features you had in mind in a single commit (including the p_upper_range option)? Also without all the previous commits for the features added in 2.6.2? Hard to review the code like this.

Trekoid:
Yes, I will create a new pull request. Wrapping projects before year end have had me tied up.

keithf4:
v3.01 has been released and had quite a lot of internal code changes. If you'd still like to have this patch considered for inclusion, please see about consolidating the changes into a single commit as well as making sure they still work with the new major version. I've also changed pg_partman to automatically determine the partitioning type based on the control column's data type. So you may be able to reuse the one internal function I created (`check_control_type.sql`).

Trekoid:
@keithf4 I will close this pull request and submit a new one based upon the current versioning. I have kept it open as a reminder to come back to this. I expect to be able to get to this next week.

keithf4:
Ok. FYI, you may want to evaluate this against the native partitioning coming up in PG 10 as well. I believe its bounds are inclusive of the lower boundary and exclusive of the top.

Trekoid:
I already have the attached working in 9.4 and 9.6 environments at work for the 2.6.2 version. I'm not sure when we will upgrade to 10 as it is taking forever for us to upgrade to 9.6.

The attached has been reworked and resubmitted against the 3.x changes in #183

PR #151, opened 2016-12-04
table_privs using information_schema.table_privileges instead of pg_authid

matheusespanhol:
Changing view table_privs in order to avoid permission problems on pg_authid table. It makes possible to use pg_partman in Amazon RDS databases.

In Amazon RDS the user available doesn't have permissions to view pg_authid catalog table. 

This change leverage the view information_schema.table_privileges.


keithf4:
Thanks! Will review and try to get in the next release.

keithf4:
If you look at the comment for this view in my code, I was not using information schema due it being a pretty significant performance bottleneck. I'd actually been using information_schema before, but I had to change away from it since the view definition does not do proper type casting to take advantage of system catalog indexes.

https://github.com/keithf4/pg_partman/issues/139

This is unfortunately another reason why I cannot support Amazon RDS with this extension.

PR #153, opened 2016-12-09
fix seq_scan on time-based partitioned tables when epoch used

dturon:
fix seq_scan on time-based partitioned tables when epoch used, probably is index only on epoch control column, not on to_timestamp(control_column). So max(to_timestamp(control_column) => to_timestamp(max(control_column)) works better:)

```
SELECT * FROM partman.run_maintenance();
^CCancel request sent
ERROR:  canceling statement due to user request
KONTEXT:  SQL statement "SELECT max(to_timestamp(utime))::text FROM test.test_p2016_12_08"
PL/pgSQL function partman.run_maintenance(text,boolean,boolean,boolean) line 142 at EXECUTE
Čas: 45869,189 ms

EXPLAIN SELECT max(to_timestamp(utime))::text FROM test.test_p2016_12_08;                                                                                                                      
                                        QUERY PLAN                                         
-------------------------------------------------------------------------------------------
 Aggregate  (cost=1568833.25..1568833.26 rows=1 width=32)
   ->  Seq Scan on test_p2016_12_08  (cost=0.00..1167331.14 rows=53533614 width=4)
```

### DIFF
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -127,8 +127,8 @@ LOOP
     AND tablename = split_part(v_row.parent_table, '.', 2)::name;
 
     v_partition_expression := case
-        when v_row.epoch = true then format('to_timestamp(%I)', v_row.control)
-        else format('%I', v_row.control)
+        when v_row.epoch = true then format('to_timestamp(max(%I))', v_row.control)
+        else format('max(%I)', v_row.control)
     end;
     IF p_debug THEN
         RAISE NOTICE 'run_maint: v_partition_expression: %', v_partition_expression;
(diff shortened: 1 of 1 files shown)


keithf4:
Thanks! Will get this included in next release

keithf4:
Actually... this works for run_maintenance, but other functions like undo_partition_time() that would need this to use the v_partition_expression value in places where the min/max is not used.

I actually have a note in the documentation for epoch partitioning to include the functional to_timestamp() index since it's required elsewhere for pg_partman to work effectively anyway. So, if anyone is having this issue, they should make that index on the parent so all children get it.

See the section in the create_parent() function that discusses the epoch flag - https://github.com/keithf4/pg_partman/blob/master/doc/pg_partaman.md#creation-functions

dturon:
Ok, but in our case we have index on epoch. In my opinion works on undo too...need test. We have large tables and extra index waste lot of space. I miss documentation:) Thx. Reject pull request i look on undo too.

PR #169, opened 2017-03-02
Don't Check for max value if infinite_time_partitions set

davidkmoat:
No need to do the extra work of checking for a maximum value if infinite_time_partitions is true. We've had some trouble with run_maintenance jobs backing up and locking due to checking a large table for this max value.

Would love your thoughts on whether this might be more efficient and/or if it causes side effects I'm   missing.

Thanks.


keithf4:
Great idea! I'll include this fix in the next update I'm working on. Thanks!

davidkmoat:
Awesome! Thanks!

keithf4:
v3.01 has been released and fixed this issue. It also contains quite a lot of other changes, so please review the CHANGELOG before upgrading. Thanks!

Issue #152, opened 2016-12-07
Amazon RDS compatibility with PG_PARTMAN 2.6.2


thomasboussekey:
Hello,

I'm using pg_partman on 2 type of `postgresql` instances:
* On premise linux servers
* Amazon RDS postgresql instances.

On my last installation attempt on an `Amazon RDS postgresql` instances, I tried to use `PG_PARTMAN` version **2.6.2**.
I'm using the sql script (ex:`/usr/pgsql-9.5/share/extension/pg_partman--2.6.2.sql`) contained in the extension replaction `@extschema@` by `partman`, my partitioning management schema.

It generated errors like this one:
```shell
ERROR: permission denied for relation pg_auhtid
```
My previous installation was performed successfully with version **2.5.0**.
I performed a `DIFF` check between the SQL scripts of versions **2.5.0** and **2.6.2**.

```shell
diff /usr/pgsql-9.5/share/extension/pg_partman--2.5.0.sql /usr/pgsql-9.5/share/extension/pg_partman--2.6.2.sql | grep authid
>          pg_authid u_grantor,
>            SELECT oid, rolname FROM pg_authid
```

I'm reinstalling with version **2.5.0**.

It seems that the pg_authid is hard to access or doesn't exist on RDS instances...

Is there a workaround to keep `PG_PARTMAN` working with `Amazon RDS postgresql` instances?

Regards,
Thomas

keithf4:
There is a pending push request to fix the pg_authid permissions issue (https://github.com/keithf4/pg_partman/pull/151). It will hopefully be in the next release.

thomasboussekey:
Thank you Keith for the update!

Have a nice day,
Thomas

keithf4:
Please see update in Issue 151. Right now I cannot support a change to using information schema due to pretty significant performance bottlenecks. Please contact AWS support about this if you really need pg_partman to work in RDS.

thomasboussekey:
Hello Keith,

I've compiled the commit on a production database, managing hundreds of
partitions.

No performance bottleneck encountered for the moment.
But we will onle begin a test campaign on Monday.

Many thanks for the information,
I keep in touch with you to inform you on our tests.

Regards,
Thomas

2016-12-16 20:35 GMT+01:00 Keith Fiske <notifications@github.com>:

keithf4:
[...]

PR #177, opened 2017-04-03
SQL evolution listed on issue #152, for @extschema@

thomasboussekey:
Hello Keith,

Here is my little changelog:

I only changed SQL DDL scripts

* Add **@extschema@**
  * For CREATE TYPE: One occurence
  * For CREATE TABLE: Three occurences
  * For CREATE FUNCTION
  * On function call **pg_catalog.pg_extension_config_dump**
   * Into index creation for table custom_time_partitions

I didn't modify `CREATE FUNCTION` by `CREATE OR REPLACE FUNCTION`.

Thomas

### DIFF
--- a/sql/functions/apply_cluster.sql
+++ b/sql/functions/apply_cluster.sql
@@ -2,7 +2,7 @@
 * Function to apply cluster from parent to child table
 * Adapted from code fork by https://github.com/dturon/pg_partman
 */
-CREATE FUNCTION apply_cluster(p_parent_schema text, p_parent_tablename text, p_child_schema text, p_child_tablename text) RETURNS void
+CREATE FUNCTION @extschema@.apply_cluster(p_parent_schema text, p_parent_tablename text, p_child_schema text, p_child_tablename text) RETURNS void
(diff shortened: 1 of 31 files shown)


keithf4:
I may see about adding the macros across the function definitions more. However, as I said before, I don't plan on supporting pg_partman for RDS until it can actually be installed as an extension.

keithf4:
Next time I do an update that touches a lot of functions, I'll see about getting these explicit macro calls in.

keithf4:
Looking at getting all the schema macros in with the next major version I'm currently working on now. With the recent search_path security issue, it's probably for the best anyway. 

https://wiki.postgresql.org/wiki/A_Guide_to_CVE-2018-1058:_Protect_Your_Search_Path

keithf4:
All of the function definitions and calls should be using the @extschema@ macro now, so doing a find-n-replace to be able to install this into RDS may work now.

Note that I still can't really support RDS since it's not installed as an extension. How you manage installing updates and tracking the version installed after that is up to you.

PR #181, opened 2017-04-20
Switch check_version to use numeric comparisons

Trekoid:
There is an easier way to do the version comparison logic. It is possible to eliminate the function completely, but I wasn't sure if you would want to go that far.


keithf4:
Nice! Didn't realize you could cast it like that. Guessing eliminating it completely would just be doing a numeric comparison right away instead of calling the function?

Trekoid:
Yes, exactly. I can roll this up with some other work I am working on and will be submitting soon.

Trekoid:
Included with #183

PR #182, opened 2017-04-21
Replace type check functions with PostgreSQL enumerated types

Trekoid:
@keithf4 , one more thing I wanted to run past you that I can bundle up with other work.
I've created enumerated types (https://www.postgresql.org/docs/9.1/static/datatype-enum.html) for the parameters where you want to limit the inputs. This accomplishes a few things:

1. Types are created before tables, solving the constraint dependency issue.
2. PostgreSQL will check the values and throw an error for you. No need for the trigger functions.
3. Cleans up the project structure, so that functions are not hiding in the tables.sql file.
4. PostgreSQL will parse the text strings as literals of the enumerated type, i.e. auto-cast.

I have run the test suite and everything still passes. See [pgpartman_test_results.txt](https://github.com/keithf4/pg_partman/files/948221/pgpartman_test_results.txt) for the results.


keithf4:
Appreciate the work, but I actually purposely avoid enums due to how difficult they are to manipulate, especially removing a value. Ran into some really hairy issues in the past with this when I needed to change a constraint value that used an enum. 

https://github.com/keithf4/pg_partman/commit/d072bbbde80107de708fed9e90ab266532a3a901
https://github.com/keithf4/pg_partman/commit/8c6830019cd3af397113aa389f0a0ba6d523b40b

It's more complicated, yes, but using the functions with the constraint makes it more flexible to be able to change the values that are allowed in a constraint. This is especially critical in an extension where being able to predictably modify things is (I think) a requirement.

Trekoid:
I agree removing values from an enumeration is a pain. The way I handle it is to create a second enum, convert the columns of any tables, drop the old enum, rename the new enum. Converting the columns of any tables can be a lot of work if the enumeration is used in many places.

I suppose the PostgreSQL team has never implemented dropping values, as it would requiring scanning the dependent tables to make sure the value isn't being used. Something that would require a transaction, and most enumeration modification commands have to be run non-transactionally. Not sure why that is the case. Probably, because it is modifying the catalog directly.

I'll leave this out of the range type support pull request.

PR #183, opened 2017-04-24
Ranged types

Trekoid:
Add partitioning over the lower or upper boundary of the ranged time data types tstzrange, tsrange, and daterange.
Added test suite in test/test_ranged. `pg_prove -ovf --dbname partman test/test_ranged/*.sql`
Includes check_version function elimination from #181.

Needs updates to the documentation.
Needs bump to version and an update script.


keithf4:
I've implemented the simpler version checking from this pull request in version 3.0.2. Was a much easier fix for it not working in 10beta1 than trying to fix my function. 

Unfortunately, I don't think I'm going to implement this ranged option at this time. With native partitioning coming to 10, I'm now trying to keep things on a track that will have features that are mostly compatible with the native options. And possibly allow the removal of a plpgsql trigger all together. While native is still missing some features (auto-inherit indexes, foreign keys & privileges), those are things on the table to be included in the future and/or they are not part of features that require a custom partitioning trigger.

This extension has already gotten more features in it than I ever thought it would have, which is both good and bad. While I have both native and custom partitioning as features, I'd like to keep from adding any additional ones unless they are critical for a large majority of users. I can see about revisiting this at a later time when native partitioning has more features built in and may allow more customizable options. 

Appreciate all the work!

PR #216, opened 2018-04-19
Doc update undo partition

blogh:
Hi, 

I noticed that the documentation of undo_partition lacks a mention of dropping triggers and functions. I propose to add it.

I was also quite surprise when I realized the partition where not undone in the order I thought they would. I did a bunch of drop / create on my partitions before playing with the undo function  so the "logical" order was not the same as the "creation order" of the partition. I propose to add a warning about that in the undo_partition doc.

Thx for the hard work.
Benoit.

### DIFF
--- a/doc/pg_partman.md
+++ b/doc/pg_partman.md
@@ -356,6 +356,7 @@ As a note for people that were not aware, you can name arguments in function cal
  * Does not currently work on native partition sets.
  * If you need to keep the data in your child tables after it is put into the parent, use this function. 
  * Unlike the other undo functions, data cannot be copied in batches smaller than the partition interval. Every run of the function copies an entire partition to the parent.
+ * When this function is run, the trigger on the parent table & the trigger function are immediately dropped (if they still exist). This means any further writes are done to the parent.
  * When this function is run, the **`undo_in_progress`** column in the configuration table is set to true if it was managed by pg_partman. This causes all partition creation and retention management to stop ONLY if it was managed by pg_partman.
  * If you are trying to un-partition a large amount of data automatically, it is recommended to run this function with an external script and appropriate batch settings. This will help avoid transactional locks and prevent a failure from causing an extensive rollback. See **Scripts** section for an included python script that will do this for you.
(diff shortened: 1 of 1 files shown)


keithf4:
There are specific `undo_partition_time()` and `undo_partition_id()` functions that will undo things in logical order. `undo_partition_native()` also does things in logical order. These functions only work on partitions made by pg_partman and it states this. 

`undo_partition` is made to work on any partition set, not specifically to the ones generated by pg_partman. In order to do logical order, partman would have to know the exact pattern of the partitioning names, which is why the above two functions work. In a general case for any partition set, that isn't predictable, so it can only go in whatever order the list returned from pg_inherits returns. 

I do have the trigger drop warning in the other functions, so I agree it should be in one you point out too. I will add that to be included in my next commit.

blogh:
Overall, I agree with your analysis and I think I should have used `undo_partition_time()` from the start.

I used `undo_partition` because it didn't have the mention about the triggers. I thought I could just backfill the last child into the parent without removing the triggers (I was just playing around with the tool, it's not a "need" I have). I was very surprised to see the parent growing :) afterwards.

Regarding the order of the Undo, I understand the reason but was also very surprised at first. I got my answer in the code but felt like It would be worth explaining the unpredictability a bit more.

On a different topic: Is it normal that psycopg2 is not a dependency of the pg_partman package in PGDG ? I understand it's not essential but it's surprising none the less to get this message : 

```
postgres@ulbdybbb01:~$ /usr/lib/postgresql/9.5/bin/reapply_indexes.py
Traceback (most recent call last):
  File "/usr/lib/postgresql/9.5/bin/reapply_indexes.py", line 3, in <module>
	import argparse, psycopg2, re, sys, time
ImportError: No module named psycopg2
```

keithf4:
I don't maintain any packages for postgres, so you'd have to contact the maintainers of those repositories about any dependencies.

blogh:
Ok thanks !

I'll close this since you plan to add the comments in a future commit.

ChristophBerg:
Fwiw I'll update the dependency of the Debian package.

PR #220, opened 2018-05-08
Allow to create unique index.

goodkiller:


keithf4:
Thanks for reporting this. I will review and if all looks good include it in the next release.

jnicholls:
I am running into the lack of unique index support and found this patch. Definitely would like to see this merged for the next release! Thanks.

PR #239, opened 2018-11-22
Check for epoch partioning in partition_data_proc

treecko1:


keithf4:
Thanks for the fix! I've included it in the next release I'm currently working on.

treecko1:
Thank you for looking into it. This was meant more like a pseudo bug report, but since I already made the changes.. So I wouldn't mind if you close this.

keithf4:
This fix has been added in 4.1.0. Thank you for your assistance!

Issue #230, opened 2018-09-21
Support moving data out of PG11 default partition


keithf4:
Add support to partition_data*() functions for moving data out of the PG11 default partition into the relevant child table(s).

Will involve using a temporary table to hold data since child table with a constraint covering data in the default cannot be added.

Airconaaron:
Hey any updates on this feature? Would love to help out in some way.

Currently at my company we had an issue where it turns out the run_maintenance() was never run since the beginning, so our default tables are massive.

With this sort of feature I imagine the logic would be similar for resizing the partition sizes as well.

keithf4:
Just a lack of time for the delay on this. Basic feature of this would be that it has to first move the data out to some other table (either temporary or not) before the relevant child table can be created for it. If you'd like to take a crack at adding this feature, I'd be glad to review and help.

keithf4:
This feature has been added to 4.1.0. Please let me know if it works ok for you.

mewwts:
Doesn't seem to work for me. I'm getting the issue in #247 when running `run_maintenance`

`ERROR:  updated partition constraint for default partition would be violated by some row`

keithf4:
Do you have data in your default partition? Moving it out isn't an automatic operation, you will have to run the `partition_data*()` function/procedure to do so.

mewwts:
Yes, I do. I must have set up the background worker incorrectly, because partitions were not made automatically. All the data in my default partition belongs in partitions not created. What happens when I then `CALL partition_data_proc('x')`?

keithf4:
For native partitioning, it moves all the data that would go into a single child partition for the given interval in a single transaction. Then commits and if any data is still in the default, it will move on to creating the next child and commit after each child is created until the default table is empty.

You can review additional options for `partition_data_proc()`  in the docs https://github.com/pgpartman/pg_partman/blob/master/doc/pg_partman.md#creation-functions

I'd recommend reviewing all the options available to see if any may allow your run of the function to run smoother for your environment.

mewwts:
Does it create the necessary child tables and attaches them too?

Thanks, I'm fortunate that I don't have live traffic on this instance yet so I don't mind the locks.

keithf4:
Yes. The procedure basically just calls the relevant `partition_data_time()` or `partition_data_id()` functions under the hood in a loop. So can read up on those for more details on it does.

mewwts:
Thank you @keithf4!

nobleread:
[...]

PR #254, opened 2019-04-15
Creating new native partitions whose constraints match rows in the default partition

jkatz:
Given:

```sql
CREATE EXTENSION pg_partman;
CREATE EXTENSION pgcrypto;

/** removed not nulls */
CREATE TABLE scans (
    id uuid DEFAULT gen_random_uuid(),
    scan float,
    created_at timestamptz
) PARTITION BY RANGE(created_at);

SELECT create_parent('public.scans', 'created_at', 'native', 'daily', p_debug => true);

INSERT INTO scans (scan, created_at)
SELECT random(), x
FROM generate_series(CURRENT_TIMESTAMP, CURRENT_TIMESTAMP + '30 days'::interval, '2 seconds'::interval) x;

SELECT run_maintenance();
```

The run maintenance would fail: with native partitioning, a new partition cannot be attached if the partition constraint matches rows that currently exist in the default partition.

This series of commits addresses the issue by performing the following:

1. (Whitespace cleanup in files I modified. Yay Atom!)
1. Add a `create_partition_default` function. This functions creates a default partition using native partitioning that is available for PostgreSQL 11+. This also modifies the `create_parent` code to create the default partition right after the new parent table is inserted into the partition config. This allows the creation of the initial partitions to succeed
1. Move any rows from the default partition that match the constraints of a newly created partition into said partition prior to attaching to the parent table. This ensures that `run_maintenance` will succeed with native partitioning for the above case.

This patch lacks updated documentation in pg_partman.md for the `create_partition_default` function as well as pgtap tests, but willing to add if needed.

### DIFF
--- a/sql/functions/create_parent.sql
+++ b/sql/functions/create_parent.sql
@@ -456,6 +456,8 @@ IF v_control_type = 'time' OR (v_control_type = 'id' AND p_epoch <> 'none') THEN
 
     RAISE DEBUG 'create_parent: v_partition_time_array: %', v_partition_time_array;
 
+    PERFORM @extschema@.create_partition_default(p_parent_table, v_job_id);
+
     v_last_partition_created := @extschema@.create_partition_time(p_parent_table, v_partition_time_array, false);
 
     IF v_last_partition_created = false THEN
@@ -606,6 +608,8 @@ IF v_control_type = 'id' AND p_epoch = 'none' THEN
         , v_template_schema||'.'||v_template_tablename
         , p_publications);
 
+    PERFORM @extschema@.create_partition_default(p_parent_table, v_job_id);
+
     v_last_partition_created := @extschema@.create_partition_id(p_parent_table, v_partition_id_array, false);
 
     IF v_last_partition_created = false THEN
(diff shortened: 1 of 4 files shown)


keithf4:
I already had this feature in development. See https://github.com/pgpartman/pg_partman/issues/230

I just baked it into the partition_data_*() functions so it's more easily usable in the rest of the extension where those functions are already used.. Will still look over your PR and see if there's anything I didn't think of, though. Thanks!

jkatz:
Cool :smile:

In the patch I proposed, the actual move is baked into the `partition_data_*` functions as well (f75bb27), but, at least in my version, to get it to work appropriately I needed to move the point in the code where the default partitions were created (5caa5b6).

keithf4:
So if I'm looking at it right, it seems you're trying to move the data out of the default partition at the time create_partition* is called? If so, this would essentially try and move data out during the maintenance run which could run into some really bad situations if someone did a large batch of inserts that went to the default. Also, when moving data out of the default, you have to first move the data somewhere else before you can actually create the child table. Postgres won't let you create it otherwise. 

I'm not seeing where you modified the partition_data*() functions either.

But, I think I've already covered most of what you did with I did so far. I changed the `check_parent()` function to a `check_default()` function which now checks both the parent and default partitions for data. So monitoring needs to run that function to keep an eye on things. And the partition_data*() functions first move the data to a temp table before creating the child tables to account for the constraint conflict that postgres will return if you try and create a child table that matches data in the default.

Moving the data out of the default as part of normal maintenance unfortunately just isn't something that would scale well. It's better to monitor for partition creation failure and handle it then.

jkatz:
@keithf4 

1. Oops, I thought I did the partition_data functions, it was the create_... functions. My bad.

2. Yeah the explanation makes sense, I agree that doing the copy at `run_maintenance` time could cause some bad situations esp. if it's moving a lot of data. I agree with the suggested approach.

I'll close out this PR. thanks!

PR #268, opened 2019-08-15
typo/whitespace

mattp-:
noticed a typo (run_mainteanance) in the doc. cleaned up some whitespace while I was there. thanks for the top notch extension 👍


keithf4:
Thanks!

Issue #273, opened 2019-09-26
Errors in partition_data_proc for native-partitioned table


nobleread:
I've experienced a problem while calling **partition_data_proc** for time-based partitioned table under partman's control (partitioned natively). Issue was that after adding a single future record, i.e. a record out of retention/premake range, to a (default) table and calling this procedure, I got stuck in a never ending loop in **partition_data_time** function after `ELSIF v_partition_type IN ('time-custom', 'native')` statement - loop's exit condition could never be satisfied and only changing `IF v_start_control > v_max_partition_timestamp THEN` to `IF v_start_control >= v_max_partition_timestamp THEN`, so changing comparison from `>` to `>=`, allowed for whole procedure to exit and do its job. Is this a mistake, that should be fixed in partman's source code?

Second thing I've noticed is that after this record was moved from default table to a proper partition, it seems like run_maintenance does not work properly any longer after part_config.premake param for this table was changed.

Please take a look how I got both errors below.

So I start off with the table configured like this:
```
-[ RECORD 1 ]--------------+--------------------------------------
parent_table               | public.foobar
control                    | datecolumn
partition_type             | native
partition_interval         | 1 day
constraint_cols            |
premake                    | 14
optimize_trigger           | 4
optimize_constraint        | 30
epoch                      | none
inherit_fk                 | t
retention                  | 37 days
retention_schema           |
retention_keep_table       | f
retention_keep_index       | f
infinite_time_partitions   | t
datetime_string            | YYYY_MM_DD
automatic_maintenance      | on
jobmon                     | f
sub_partition_set_full     | f
undo_in_progress           | f
trigger_exception_handling | f
upsert                     |
trigger_return_null        | t
template_table             | partman.template_public_foobar
publications               |
inherit_privileges         | f
constraint_valid           | t
```

Current time is:
```
db=# select now();               
              now                
-------------------------------- 
 09/26/2019 08:29:57.510308 EDT  
(1 row)                     
```     

Adding a 'future' record:
`db=# insert into foobar values (..., '11/27/2019');
`

And confirm record was put into default table
```
db=# select * from foobar_default;
... | datecolumn
...-+-----------
... | 11/27/2019

```
Call partition_data_proc with a proposed quickfix - everything works and new partition is created, record is moved properly. Without quickfix, I'm stuck in a never ending loop
```
db=# \d+ foobar
...
foobar_p2019_10_09 FOR VALUES FROM ('10/09/2019') TO ('10/10/2019'),
[...]

PR #277, opened 2019-10-03
v4.2.1 Fix bug in partition_data_time causing never-ending loop while…

nobleread:
v4.2.1 Fix bug in partition_data_time causing never-ending loop while calculating partition range. Github issue #273


keithf4:
Thanks! You'll also need to do an extension update file as well. Path should be from 4.2.0 to 4.2.1.

https://github.com/pgpartman/pg_partman/tree/master/updates

nobleread:
Missed that, sorry! Now everything should be OK.

PR #278, opened 2019-10-04
Fix (unique) index inheritance from template

jcoleman:
Once a duplicate index was detected on the template table, all
subsequent indexes were being treated as duplicates since the duplicate
boolean variable wasn't being reset after each loop iteration.

The test harness didn't catch this since it actually avoided creating
duplicative indexes on template tables.

I moved around the test harness setup so that a duplicate exists and
added test assertions showing the behavior before and after the fix.

Co-authored-by: Jonathan Paul Lukens <jonathan.lukens@gmail.com>

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,6 @@
+4.2.2
+-- Fix bug in template table index inheritance that prevented any additional indexes from being copied after finding an index definition that exists on both the template and parent tables. Github Pull Request #278.
+
 4.2.1
 -- Fix bug in partition_data_time causing never-ending loop while calculating partition range for natively partitioned tables. Github issue #273
 
--- a/META.json
+++ b/META.json
@@ -1,7 +1,7 @@
 {
     "name": "pg_partman",
     "abstract": "Extension to manage partitioned tables by time or ID",
-    "version": "4.2.1",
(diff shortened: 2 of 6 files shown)


keithf4:
Thanks for this! If you want to do a full release patch like nobleread did for PR https://github.com/pgpartman/pg_partman/pull/277 I could then merge it in. Otherwise I'll include this fix in my next release and credit you in CHANGLOG.

jcoleman:
Would a full release end fairly quickly on PGDG apt repos? If so, that would be extremely helpful to us.

keithf4:
I do not control the pgdg packages for this. Don't have an ETA on my own planned next release, so if you'd like this fix in PGDG sooner, as soon as I can tag a release here, the sooner it will be picked up by the pgdg maintainers.

jcoleman:
Would you be able to tag a release here then at least?

keithf4:
If you can provide a full release with the relevant update file and other changes that previous updates require, then yes, I can merge it in and tag it. The other PR that I tagged previously should give you some guidance.

keithf4:
Thank you! Will review and try and have tagged next week.

keithf4:
Version 4.2.2 has been tagged. Thanks again for the bugfix!

PR #279, opened 2019-10-05
Add Docker test environment for easy development.

jcoleman:
Add a Dockerfile with a Postgres build environment as well as simple
scripting to make it easy for contributors to quickly iterate on
extension development without needing to configure a build environment.

Unfortunately I can't easily add the Docker setup to the Makefile (as
it's normally common to do) since the existing Makefile requires the
full Postgres extension build environment, which would defeat a lot of
the purpose of the Docker environment in the first place.

### DIFF
--- a/build_for_tests.sh
+++ b/build_for_tests.sh
@@ -1,0 +1,13 @@
+#! /bin/bash
+
+set -e
+
+cd /home/pg_partman
+
+make && make install
+
+psql -U postgres -c "DROP DATABASE IF EXISTS partman_test"
+psql -U postgres -c "CREATE DATABASE partman_test"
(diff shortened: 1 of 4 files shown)


build_for_tests.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/bash
+
```
hanefi, line 2:
```suggestion

set -e

```

I'd expect this script to halt in case `make` or `make install` fails for some reason. This also applies for the other queries down below.

calebj:
Please take a look at [my take](https://github.com/calebj/pg_partman/tree/feat-docker-tests) on this functionality and let me know if it should be a separate PR.

PR #291, opened 2020-01-04
Feature: function to dump partition definition

### LINKED ISSUE #260: Dump configuration
Are there by any chance plans to include a function that would dump configuration to `SELECT create_parent`/`UPDATE part_config`/etc. calls? Output would be useful for inclusion in a .sql schema dump instead of having pg_dump naively dump each existing partition.

jcoleman:
Many application development frameworks dump the structure of the
database after running their schema migrations to aid development. If
you apply that structure dumping (e.g., using pg_dump) naïvely to
tables partitioned by time you'll quickly either end up with a
dysfunctional schema (e.g., there won't partitions for current data) or
with unexpected perma-diffs in your repo (e.g., with the set of
partitions changing every month.

To handle this problem it's useful to exclude partitions from the
structure dumping and instead use a custom function to export the SQL
necessary to recreate the current valid set of parititions, since this
shouldn't change unless the partition configuration changes.

Here I add a function to dump a partition's configuration to SQL
statements.

Fixes #260.

This still needs documentation added, but I wanted to get this much pushed up for review as soon as possible.


keithf4:
Thanks for this! Will review when I have a chance.

keithf4:
So this is doing pretty much exactly how I thought I would do it functionally, so thanks a lot! 

I think the only things about this PR would be to adjust things to match the coding patterns I've used throughout the extension already:

- Prepend all parameters to the functions with `p_`. Eg `p_table_name`.
- Prepend all variables used in the function with `v_`. Eg `v_create_parent_definition`

I started doing the above two things myself to help avoid ambiguity between object names (tables, columns, etc) vs the variables being used. And also to differentiate between parameters being passed in vs internally used variables. Can see any of my other plpgsql functions to see how I've done it before.

The other thing I would request is for `create_parent_definition` & `update_part_config_definition`, use `format()` instead of string concatenation. Avoids security issues around user submitted values (not a huge issue here) and makes the string itself a lot easier to manage without having to worry about excessive special character escaping. It's also consistent with how I've already been managing string statement creation throughout the extension.
https://github.com/jcoleman/pg_partman/blob/add-dump-def-function/sql/functions/dump_partitioned_table_definition.sql#L129
https://github.com/jcoleman/pg_partman/blob/add-dump-def-function/sql/functions/dump_partitioned_table_definition.sql#L150

If you're able to make these changes to your PR, then I'd be able to just use your code mostly as is and include it in the next release. Otherwise I'll have to adjust things and I may not get it in the next release.

Again, thanks for the contribution!

keithf4:
Got it working and just another question... The returns between each parameter look nice to read, but doesn't this make it harder to feed back into another script or copy-n-paste to run the commands? Have to do some additional cleanup with the plus signs to get that to work. Maybe just better to return the statements as is, perhaps even as rows like I do with the `show_*` commands?
```
keith=# select * from partman.dump_partitioned_table_definition('partman_test.time_taptest_table');
              dump_partitioned_table_definition               
--------------------------------------------------------------
 SELECT partman.create_parent(                               +
         p_parent_table := 'partman_test.time_taptest_table',+
         p_control := 'col3',                                +
         p_type := 'partman',                                +
         p_interval := 'daily',                              +
         p_constraint_cols := NULL,                          +
         p_premake := 6,                                     +
         p_automatic_maintenance := 'on',                    +
         p_inherit_fk := true,                               +
         p_epoch := 'none',                                  +
         p_upsert := '',                                     +
         p_publications := NULL,                             +
         p_trigger_return_null := true,                      +
         p_template_table := NULL,                           +
         p_jobmon := true                                    +
 );                                                          +
 UPDATE partman.part_config SET                              +
         optimize_trigger = 6,                               +
         optimize_constraint = 30,                           +
         retention = NULL,                                   +
         retention_schema = NULL,                            +
         retention_keep_table = true,                        +
         retention_keep_index = true,                        +
         infinite_time_partitions = false,                   +
         datetime_string = 'YYYY_MM_DD',                     +
         sub_partition_set_full = false,                     +
         trigger_exception_handling = false,                 +
         inherit_privileges = true,                          +
         constraint_valid = true                             +
 WHERE parent_table = 'partman_test.time_taptest_table';
(1 row)
```

jcoleman:
I'm happy to adjust the naming style to match the project; I might not be able to plan that until Friday next week; would that still fit in your timeframe for the next release?

As far as the output styling, I have mixed feelings/some questions:
- For one-off testing/manual usage, having it all one line etc. is useful.
- For inclusion into a SQL file, having the line breaks and indentation matches the style of pg_dump.

Returning multiple rows (I think that was what you were referring to with the `show_*` functions?) seems to me to make it a good bit more awkward to use programmatically.

Thoughts?

keithf4:
Looking over that again, looks like it's just psql that adds those plus signs in and can be turned off by sending `-A` when running it. So learned something new there. Return format is fine like that for now, then. 

Not sure my release will even be ready by next Friday, so that's fine for now. If I need to get it out sooner, I'm fine doing another release to add this in later then.

keithf4:
I believe this function can also be marked STABLE as well. 

Also, just another code pattern, but I typically put the LANGUAGE definition at the beginning part of the function vs the end.

jcoleman:
@keithf4 I believe I've addressed all of the feedback; let me know if there's anything I missed.

keithf4:
Looks great! I'll probably move it into the `test_native` folder and may clean up a few small things just around formatting preferences. But will definitely be in the next version I'm hoping to have out in the next few weeks. Thanks!

jcoleman:
What's the reason for `test_native`? It works for both native and partisan partitioning I believe. (Unless I'm misunderstanding the name of that folder.)

keithf4:
All the tests at the top level work on all versions of PG, native or not. Any test that involves native partitioning I put into the `test_native` folder, so I can only run those tests on PG10+

jcoleman:
Ah, that makes sense, I didn't understand purpose of that directory.

keithf4:
Pushed up into the "migrate" branch if you want to test it out. Did adjust the parameter name to be consistent with other functions - https://github.com/pgpartman/pg_partman/tree/migrate

keithf4:
Testing this on PG10, the tap test you provided fails because you can't make the primary key on the parent. I just removed the primary key definition for now and test seems to still work fine and test functionality as needed.

keithf4:
This has been pulled in as of version 4.3.0

PR #304, opened 2020-04-12
Comparasion between timestamptz and null is always null

### LINKED ISSUE #303: Not new partitions are created if partitions are empty and default have values
Having this config:
```
    parent_table    | partition_type | partition_interval | control | premake | undo_in_progress | sub_partition_set_full |  epoch  | infinite_time_partitions | retention
--------------------+----------------+--------------------+---------+---------+------------------+------------------------+---------+--------------------------+-----------
 public.history_str | native         | 01:00:00           | clock   |      72 | f                | f                      | seconds | f                        | 7 days
```

During some time I had not inserted data into that table.
Partitions were dropped (except one, not deleted by partman).
Then some data start to be inserted into history_str and now new partitions are being created.

The run_maintenance in debug mode shows this line:
```
NOTICE:  run_maint: v_current_partition_timestamp: <NULL>, v_max_time_parent: 2020-04-12 11:39:32+02
```

``v_current_partition_timestamp`` is ``NULL`` because there is no data in the uniqe partition remaining.
``v_max_time_parent`` is the latest row inserted in the default table.

The functions [compares](https://github.com/pgpartman/pg_partman/blob/258679998659f7f3d2fb75b0d870eb19ff8296f6/sql/functions/run_maintenance.sql#L230) that two values:
```sql
IF v_max_time_parent > v_current_partition_timestamp THEN
```

I think that is the problem. That comparison (time VS NULL) always return false.
I have made a small test in PG12:

```sql
create function test_compare_time_null() RETURNS void
    LANGUAGE plpgsql AS $$
DECLARE
v_current_partition_timestamp   timestamptz;
BEGIN
IF now() < v_current_partition_timestamp THEN
  RAISE NOTICE 'compare correct';
ELSE
  RAISE NOTICE 'compare false';
END IF;

IF now() > v_current_partition_timestamp THEN
  RAISE NOTICE 'compare correct';
ELSE
  RAISE NOTICE 'compare false';
END IF;
END
$$;
```

```
# select test_compare_time_null();
NOTICE:  compare false
NOTICE:  compare false
```

Meanwhile I will use ``infinite_time_partitions``.

adrianlzt:
If v_current_partition_timestamp is NULL (there is no data in the
partitioned tables), the comparison is always false, even if the
default table have data.
This error avoid setting the correct value for v_current_partition_timestamp
avoiding the creation of new partitions.

This query returns true:
SELECT ('2018-03-11 02:30'::timestamp > NULL) IS NULL;

Fixes #303

### DIFF
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -227,7 +227,7 @@ LOOP
         IF p_debug THEN
             RAISE NOTICE 'run_maint: v_current_partition_timestamp: %, v_max_time_parent: %', v_current_partition_timestamp, v_max_time_parent;
         END IF;
-        IF v_max_time_parent > v_current_partition_timestamp THEN
+        IF v_current_partition_timestamp IS NULL OR v_max_time_parent > v_current_partition_timestamp THEN
             SELECT suffix_timestamp INTO v_current_partition_timestamp FROM @extschema@.show_partition_name(v_row.parent_table, v_max_time_parent::text);
         END IF;
 


adrianlzt:
```
Ordinary comparison operators yield null (signifying “unknown”), not true or false, when either input is null
```
https://www.postgresql.org/docs/current/functions-comparison.html

adrianlzt:
Modified by @keithf4 . See #303

Issue #300, opened 2020-03-06
inherit_template_properties PG10 inheritance issue


KevMurph:
https://github.com/pgpartman/pg_partman/blob/258679998659f7f3d2fb75b0d870eb19ff8296f6/sql/functions/inherit_template_properties.sql#L156

The above line has given me some problems in the past on PG10, but I've been able to resolve it by swapping it out the section of code with something like below. I believe it has something to do with having having indexes on a separate tablespace from the parent table / partitions.

ELSE
-- statement column should be just the portion of the index definition that defines what it actually is
    v_sql := format('CREATE %s INDEX ON %I.%I %s', CASE WHEN v_index_list.indisunique = TRUE THEN 'UNIQUE' ELSE '' END, v_child_schema, v_child_tablename,v_index_list.statement);
    IF v_index_list.tablespace_name IS NOT NULL 
THEN
       -- v_sql := v_sql || format(' TABLESPACE %I', v_index_list.tablespace_name);
       execute format('set default_tablespace = %s',  v_index_list.tablespace_name );
    END IF;
    RAISE NOTICE 'Create index: %', v_sql;
    EXECUTE v_sql;
    reset default_tablespace;
END IF;

Here is the error message:
db=# SELECT partman.run_maintenance(p_analyze:=false);
ERROR:  syntax error at or near "TABLESPACE"
CONTEXT: PL/pgSQL function inherit_template_properties(text,text,text) line 156 at EXECUTE
SQL statement "SELECT partman.inherit_template_properties(p_parent_table, v_parent_schema, v_partition_name)"
PL/pgSQL function create_partition_time(text,timestamp with time zone[],boolean,boolean) line 228 at PERFORM
PL/pgSQL function partman.run_maintenance(text,boolean,boolean,boolean) line 280 at assignment
DETAIL:
HINT:
CONTEXT: PL/pgSQL function create_partition_time(text,timestamp with time zone[],boolean,boolean) line 459 at RAISE
PL/pgSQL function partman.run_maintenance(text,boolean,boolean,boolean) line 280 at assignment
DETAIL:
HINT:
CONTEXT:  PL/pgSQL function partman.run_maintenance(text,boolean,boolean,boolean) line 398 at RAISE

keithf4:
I'm having some trouble reproducing this error. This TAP test 

https://github.com/pgpartman/pg_partman/blob/master/test/test_native/tablespace/test-time-daily-native-tablespace-template.sql

[...]

PR #313, opened 2020-05-11
BUGFIX: Certain index statements lead to invalid generated SQL in inherit_template_properties

KevMurph:
This patch resolved my issue (#300). It seems that certain valid index definitions can lead to malformed dynamic SQL in this area of code. Bypassing this syntax and temporarily setting the default_tablespace seems to sidestep the issue.

### DIFF
--- a/sql/functions/inherit_template_properties.sql
+++ b/sql/functions/inherit_template_properties.sql
@@ -153,11 +153,12 @@ IF current_setting('server_version_num')::int >= 100000 THEN
             -- statement column should be just the portion of the index definition that defines what it actually is
             v_sql := format('CREATE %s INDEX ON %I.%I %s', CASE WHEN v_index_list.indisunique = TRUE THEN 'UNIQUE' ELSE '' END, v_child_schema, v_child_tablename, v_index_list.statement);
             IF v_index_list.tablespace_name IS NOT NULL THEN
-                v_sql := v_sql || format(' TABLESPACE %I', v_index_list.tablespace_name);
+                execute format('SET LOCAL default_tablespace = %s',  v_index_list.tablespace_name);
             END IF;
 
             RAISE DEBUG 'Create index: %', v_sql;
             EXECUTE v_sql;
+            RESET default_tablespace;
 
         END IF;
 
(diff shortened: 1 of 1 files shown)


KevMurph:
This patch resolved my issue (https://github.com/pgpartman/pg_partman/issues/300). I believe that certain, valid index definitions can lead to malformed dynamic SQL in this area of code.

sql/functions/inherit_template_properties.sql
```diff
@@ -153,11 +153,12 @@ IF current_setting('server_version_num')::int >= 100000 THEN
             -- statement column should be just the portion of the index definition that defines what it actually is
             v_sql := format('CREATE %s INDEX ON %I.%I %s', CASE WHEN v_index_list.indisunique = TRUE THEN 'UNIQUE' ELSE '' END, v_child_schema, v_child_tablename, v_index_list.statement);
             IF v_index_list.tablespace_name IS NOT NULL THEN
-                v_sql := v_sql || format(' TABLESPACE %I', v_index_list.tablespace_name);
+                execute format('SET LOCAL default_tablespace = %s',  v_index_list.tablespace_name);
```
keithf4, line 156: Should leave this as an identity (%I) in the format command

keithf4: Actually sorry, looks like %I doesn't work here. Will check it out later. Thanks!

KevMurph:
Could you elaborate further on how this doesn't work? 
I installed PGTAP and ran through the tests with and without my patch and got the same number of failures (3) for both.

Test Summary Report
-------------------
./test-time-weekly-daily-subpart.sql     (Wstat: 0 Tests: 277 Failed: 3)
  Failed tests:  169-170, 173
Files=33, Tests=7540, 57 wallclock secs ( 1.01 usr  0.16 sys +  0.35 cusr  0.38 csys =  1.90 CPU)
Result: FAIL

KevMurph:
I believe that what we're looking for is '%L' since it's treated like a literal in the command. It's true that the tablespace has to follow standard naming conventions for postgres, but that is accounted for when the tablespace is set on parent / template.

keithf4:
Is there any chance you could provide an example that causes this issue? I'm really hesitant on introducing any bug fix for which I'm unable to reproduce the issue. I prefer to be able to cover that bug via unit tests so I know it's handled via any future changes as well.

KevMurph:
Updated Issue #300

PR #315, opened 2020-05-19
Fix "partiton" Typos

dtseiler:
A few cases of "partition" being typed as "partiton". First noticed in docs but it's in quite a few scripts as well as both comments and output.


keithf4:
Well that's embarrassing. I'll review and get that in when I have a chance.

PR #348, opened 2021-03-26
Inherit reloptions

erthalion:
Hi,

Just faced an interesting situation when reltoptions were not propagated on
children tables. From what I see it's not supported via native partitioning,
but maybe possible to inherit them from a template table. This PR is the most
straightforward change that has worked for particular case I have in mind, so I
wanted to ask if it makes sense or not?

### DIFF
--- a/sql/functions/inherit_template_properties.sql
+++ b/sql/functions/inherit_template_properties.sql
@@ -11,6 +11,7 @@ v_dupe_found            boolean := false;
 v_fk_list               record;
 v_index_list            record;
 v_inherit_fk            boolean;
+v_relopt                record;
 v_parent_index_list     record;
 v_parent_oid            oid;
 v_parent_table          text;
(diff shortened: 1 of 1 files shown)


keithf4:
Interesting find. Will review it when I have a chance, but I would request that this also be submitted as a bug to the PostgreSQL community. Long term, this should be handled in core.

https://www.postgresql.org/support/

erthalion:
> I would request that this also be submitted as a bug to the PostgreSQL community. Long term, this should be handled in core.

For sure. I'm surprised myself that it's not handled native, but it doesn't seem to be a bug, `partitioned_table_reloptions` just says

```
/*
 * There are no options for partitioned tables yet, but this is able to do
 * some validation.
 */
```

I couldn't find yet any discussion in hackers explaining why, but at the very least there are already some [patches][1] in this direction.

[1]: https://www.postgresql.org/message-id/flat/95b1dd96-8634-4545-b1de-e2ac779beb44%40www.fastmail.com

leandrotocachelis:
My partition table have custom autovacuum values and I need set manually every week.
For me it's a good idea, it makes sense.

keithf4:
This PR has been incorporated into the 4.5.1 release. Thank you!

PR #371, opened 2021-07-05
fix(create_parent): don't date_trunc when p_start_partition is set

### LINKED ISSUE #369: p_start_partition not honored (always date_trunc)
When using a p_start_partition of '04-07-2021' and a p_interval of '9 weeks' it'll result in INVALID time partitions being created. This because all the start dates of the partitions are truncated to the first of the month. This creates overlapping partitions if you're using an interval that's not standard.

```
+--------------------------+--------------------------+--------------+------------+-------+
|child_start_time          |child_end_time            |child_start_id|child_end_id|suffix |
+--------------------------+--------------------------+--------------+------------+-------+
|2021-07-01 00:00:00.000000|2021-09-02 00:00:00.000000|NULL          |NULL        |2021_07|
|2021-09-01 00:00:00.000000|2021-11-03 00:00:00.000000|NULL          |NULL        |2021_09|
+--------------------------+--------------------------+--------------+------------+-------+
```

I think it'd be best if we put the date_trunc behind a flag, or don't do it if p_start_partition is passed in.

wesselvdv:
Resolves #369

### DIFF
--- a/sql/functions/create_parent.sql
+++ b/sql/functions/create_parent.sql
@@ -387,48 +387,49 @@ IF v_control_type = 'time' OR (v_control_type = 'id' AND p_epoch <> 'none') THEN
             END IF;
     END CASE;
 
-   -- First partition is either the min premake or p_start_partition
+    -- First partition is either the min premake or p_start_partition
     v_start_time := COALESCE(p_start_partition::timestamptz, CURRENT_TIMESTAMP - (v_time_interval * p_premake));
+    v_base_timestamp := p_start_partition::TIMESTAMPTZ;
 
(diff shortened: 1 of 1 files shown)


keithf4:
This is great! Thanks you!

Once we get the questions answered on your other PR, it would be great if you could work these two PRs into a single extension update for pg_partman. You can look at examples in the `/updates` folder. This would just be a patch release, so just increment the final, third number one. 

Would also be good to have a pgtap test to ensure this works for the foreseeable future despite any other updates.

If that's done, then I can likely just merge the PR.

PR #391, opened 2021-09-28
Adding CREATE INDEX CONCURRENTLY to avoid lock in the online scenario 

yahavb:
In the case of a large table (billions of rows), creating an index may acquire an exclusive lock on the original table hence adding CONCURRENTLY


keithf4:
Thank you!

PR #441, opened 2022-02-17
Test that days retention works with weekly partitions

jcoleman:
It's currently possible to configured N days of retention (in non-week
sized increments) on a weekly partitioned table, and, indeed, it  already
works with the current implementation in `drop_partition_time()`, but given
it's not immediately obvious whether or not it will work it seems worth
putting it under test and documenting it explicitly.


keithf4:
Thanks for this!

PR #474, opened 2022-07-14
Fix run_maintenace when no default table exists

sj26:
We're encountering an error like this when running `CALL run_maintenance_proc()` on our database with a native range partitioned table without partitions:

```sql
psql:.../db/structure.sql:8965: ERROR:  null values cannot be formatted as an SQL identifier
CONTEXT: PL/pgSQL function partman.run_maintenance(text,boolean,boolean) line 323 at EXECUTE
SQL statement "SELECT partman.run_maintenance('public.agent_details', p_jobmon := 't')"
PL/pgSQL function partman.run_maintenance_proc(integer,boolean,boolean) line 42 at EXECUTE
DETAIL:
HINT:
CONTEXT:  PL/pgSQL function partman.run_maintenance(text,boolean,boolean) line 410 at RAISE
SQL statement "SELECT partman.run_maintenance('public.agent_details', p_jobmon := 't')"
PL/pgSQL function partman.run_maintenance_proc(integer,boolean,boolean) line 42 at EXECUTE
```

We're using v4.5.1. I traced it to:

https://github.com/pgpartman/pg_partman/blob/v4.5.1/sql/functions/run_maintenance.sql#L317

`v_default_tablename` is `NULL`.

That should be set to the parent tablename earlier:
https://github.com/pgpartman/pg_partman/blob/v4.5.1/sql/functions/run_maintenance.sql#L168

but `v_is_default` will be `NULL` if there are no partitions at all, and `(NULL != 'DEFAULT') IS NULL`, and `IF NULL THEN ... END` will not enter the conditional block because `NULL` is not `TRUE`, so it doesn't enter the conditional and default to the parent table name.

`(NULL IS DISTINCT FROM 'DEFAULT') IS TRUE`, however, and works the way it should, and preserves `('any string' IS DISTINCT FROM 'DEFAULT') IS TRUE`, while `('DEFAULT' IS DISTINCT FROM 'DEFAULT') IS FALSE`.

Sorry I'm not familiar enough with the test suite to write a test. If someone can provide some pointers I'll have a go. I presume it would require adding a test in `test/test-id-run-maint.sql`?

### DIFF
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -169,7 +169,7 @@ LOOP
         WHERE n.nspname = v_parent_schema
         AND c.relname = v_default_tablename;
 
-        IF v_is_default != 'DEFAULT' THEN
+        IF v_is_default IS DISTINCT FROM 'DEFAULT' THEN
             v_default_tablename := v_parent_tablename;
         END IF;
     ELSE


keithf4:
Just to clarify, you are saying there are no child tables at all in this partition set?

sj26:
Yes, that’s right.

We remove the default partition so that it doesn’t accumulate unpartitioned data which becomes a problem to partition later. In continuous integration environments with a fresh database with no default partition this means there are no child tables until the maintenance process is run, for example.

keithf4:
Ok. I wanted to make sure. pg_partman cannot work without having at least one child table. Otherwise it has no idea what the "next" table to create is. The create_parent() function should create at least the premake child tables for that. And retention should never drop the last child table. So how are you getting in the situation where there are no child tables?

sj26:
We do a schema dump and restore between continuous integration test runs, and ignore child partitions because they make the schema unstable.

How does create_parent know what the "next" table is?

sj26:
Sorry, we might be holding it wrong! How would you recommend make a stable, repeatably loadable schema for setting up ephemeral test environments? We're using Ruby on Rails and so have decorated it's schema dumper to add some partman commands at the end to try and get things consistently configured.

keithf4:
Another option would be to use the `dump_partitioned_table_definition()` on the original system to generate the `create_parent()` call along with an `UPDATE` statement to restore the remaining configuration options. Note this currently doesn't account for sub-partitioning. You could then just run those statements in your test environment vs restoring the config table contents.

keithf4:
Just checking to see if any of the suggestions i made worked for you? 

I won't be merging this PR at this time.

sj26:
No worries, we'll find another solution.

> Another option would be to use the dump_partitioned_table_definition() on the original system to generate the create_parent() call along with an UPDATE statement to restore the remaining configuration options. Note this currently doesn't account for sub-partitioning. You could then just run those statements in your test environment vs restoring the config table contents.

This sounds very promising. Perhaps we can teach the rails schema dumper to do something like this.

PR #499, opened 2023-01-03
Remove PG11+ guard on retention_keep_index

robertdecaire:
While it's true that you can't drop child indexes on native partitions in PG 11+, once you've detached the partition it's no longer a child index and can be dropped. We've been having to do this via a separate script since `retention_keep_index = false` does nothing if this guard is in place. It appears to work fine when this is removed, however.

### DIFF
--- a/sql/functions/drop_partition_time.sql
+++ b/sql/functions/drop_partition_time.sql
@@ -217,43 +217,39 @@ LOOP
                     PERFORM update_step(v_step_id, 'OK', 'Done');
                 END IF;
             ELSIF v_retention_keep_index = false THEN
-                IF v_partition_type = 'partman' OR 
-                       ( v_partition_type = 'native' AND  current_setting('server_version_num')::int < 110000) THEN
-                    -- Cannot drop child indexes on native partition sets in PG11+
-                    FOR v_index IN 
-                        WITH child_info AS (
-                            SELECT c1.oid
-                            FROM pg_catalog.pg_class c1
-                            JOIN pg_catalog.pg_namespace n1 ON c1.relnamespace = n1.oid
-                            WHERE c1.relname = v_row.partition_tablename::name
-                            AND n1.nspname = v_row.partition_schemaname::name
-                        )
-                        SELECT c.relname as name
-                            , con.conname
-                        FROM pg_catalog.pg_index i
-                        JOIN pg_catalog.pg_class c ON i.indexrelid = c.oid
-                        LEFT JOIN pg_catalog.pg_constraint con ON i.indexrelid = con.conindid
-                        JOIN child_info ON i.indrelid = child_info.oid
-                    LOOP
-                        IF v_jobmon_schema IS NOT NULL THEN
-                            v_step_id := add_step(v_job_id, format('Drop index %s from %s.%s'
-                                                                , v_index.name
-                                                                , v_row.partition_schemaname
-                                                                , v_row.partition_tablename));
-                        END IF;
-                        IF v_index.conname IS NOT NULL THEN
-                            EXECUTE format('ALTER TABLE %I.%I DROP CONSTRAINT %I'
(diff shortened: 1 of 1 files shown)


keithf4:
I'd actually been considering removing this feature of just dropping indexes in the next major release, but if you still find it a useful feature I can keep it in. Do you need this feature built in anytime soon or would waiting until the 5.x release that would incorporate this be ok? Note that it will have a minimum requirement of PG14 so the need for PG11 support will be inherently gone. Should hopefully be release by the end of 2023Q1

robertdecaire:
> I'd actually been considering removing this feature of just dropping indexes in the next major release, but if you still find it a useful feature I can keep it in. Do you need this feature built in anytime soon or would waiting until the 5.x release that would incorporate this be ok? Note that it will have a minimum requirement of PG14 so the need for PG11 support will be inherently gone. Should hopefully be release by the end of 2023Q1

We're using this feature (or trying to) extensively in our current setup. We do a lot of upserts to some very large tables which are partitioned daily, and due to the limitations on execution-time partition pruning on updates and deletes, it scans through every partition whenever we have an update, which can lead to a lot of slowdown. To alleviate this, after we're confident that we will no longer be writing new data to a partition we detach it and attach it to a "history" table with an identical schema, and join that table to the "live" table on read queries, where necessary. This minimizes the number of partitions in the table being written to. However, this sometimes results in name collisions on indexes of partitions during the detach/reattach step, due to the procedural name generation and limitation on the length of index names. So we want to just drop the indexes and let new indexes be generated when the partitions are attached to the history tables, which will necessarily have no name collisions with pre-existing indexes.

I'm led to believe that PG14 corrects the partition pruning problem and our history table workaround wouldn't be necessary, but we're on RDS Aurora and running PG13 there for the time being until we can get clearance to migrate everything to PG14, which only became available midway through last year, immediately after we upgraded to PG12 :) . Considering that you're targeting PG14 as the minimum, we wouldn't be able to benefit from this update there, and it wouldn't be necessary anyway. And of course, we're at the mercy of RDS as to if and when they decide to update their pg_partman extension.

With that in mind, I don't think this fix is likely to be useful to you. Maybe if someone has the same problem we do, they'll spot this discussion and learn how to fix it themselves. But feel free to close this PR if you don't think you'll be keeping the feature in future versions.

At any rate, since the function in question is just a routine in the `partman` schema, we just manually updated that function by running the script I altered, and that will work fine for us so long as we remember to do that every time we spin up a new DB.

keithf4:
So I will be leaving the index removal support in, but it's a bit different than your PR. Right now it only works if you're not moving the children to a different schema. So I moved it around a bit so it's supported there too.

Only thing to be aware of is that, with detaching the child table first, you cannot have ANY data in the partition set anymore that may be referred to by a foreign key. Otherwise you're not allowed to detach. This is why, if the table is just being dropped, I do not try and detach it first.

keithf4:
Version 5.0.0 has been released. Please let me know if this is working ok for you as it was implemented.

psyskeptic:
Hey guys
Also i think that this feature as retention_keep_index=false should be remain.
We have daily partitions with a lot of indexes, but after some period of time we do not want to keep data attached. So we deatach old partitions. After that we no longer need indexes on them, as they stored as archive data, dwh analytics read all data so they dont need any index.

But im curious, is there any feature to move old partitions not just to other SCHEMA, but to reatach to other table (). In that situation we would like to keep indexes.
@robertdecaire you reatach those tables by your own scripts? or it can be configured in partman somehow?

keithf4:
This feature has remained in 5.0, and has been enhanced to be able to drop the indexes on tables moved to another schema.

There are currently no features to detach and reattach to another partition set. Not doesn't really seem like a common usage and is the first I've heard it mentioned, so may be a bit out of scope for this extension at this time. If I see it requested more, I will consider it.

PR #534, opened 2023-06-18
Update README.md

RudneyEduardo:
Add possible windows installation

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,12 @@ Recommended:
  * pg_jobmon (>=v1.4.0). PG Job Monitor will automatically be used if it is installed and setup properly.
 https://github.com/omniti-labs/pg_jobmon
 
+
+
 In the directory where you downloaded pg_partman, run
 
     make install
+ 
 
 If you do not want the background worker compiled and just want the plain PL/PGSQL functions, you can run this instead:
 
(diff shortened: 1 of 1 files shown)


keithf4:
Apologies but I do not have Windows as part of my current development cycle, so I would prefer not to maintain instructions for doing so, especially since I have no way to confirm it for myself. If you'd like to maintain a Discussion with these instruction, I would not object.

https://github.com/pgpartman/pg_partman/discussions

vitaly-burovoy:
@RudneyEduardo, if Windows now has `cat` and `grep`  etc. why just not use `make install`?

RudneyEduardo:
> @RudneyEduardo, if Windows now has `cat` and `grep` etc. why just not use `make install`?

@vitaly-burovoy because i was having some troubles with make and grep, so i thought it was another way to install pg_partman if another person is having the same issue

vitaly-burovoy:
I think it is better to find a way (and document it) how to use `make` on Windows to install the extension.

Particularly you document how to install it but it lacks how to upgrade it. When you find a way to document upgrade method you will find yourself creating the Makefile.

This extension uses the Postgres's infrastructure[1] to maintain (create, copy and delete) required files and it (infrastructure) is tested under Windows and works well for many users.

I think some of other extensions have that explanation but neither the author nor I have Windows to check it.

[1] https://www.postgresql.org/docs/current/extend-pgxs.html

vitaly-burovoy:
P.S.: have you tried, for example, pgxnclient[1] as users advice[2]?
P.P.S.: again, I have not tried because I do not have Windows, but maybe it will solve your problem… 😇

[1] https://pypi.org/project/pgxnclient/
[2] https://stackoverflow.com/questions/45881944/pgxn-client-for-windows

PR #544, opened 2023-07-03
PoC migration generator

vitaly-burovoy:
I tried to play with the code but it is very hard when have some functions in a single "update" file, others - in "base" files (one per function).

Moreover it will be difficult to apply changes from those "update" file back to "base" files.

I somehow managed to write Makefile rules to have DDLs mostly in "base" files which can be merged to get mentioned "update" file.

It improves (at least for me) inspection of the code and keeps the updated migration file in a consistent state.

At this point generated "update" SQL file (to v5.0.0-beta) is almost the same as the original one (except several withespaces, comments and type casting). Other changes can be made right into "base" files and they will be automatically reflected in the "update" one.

~Unfortunately I could not solve a problem with autodetection of changes (e.g. if you change "09.cleanup_temp.sql", then `make all` does not reassemble files), a workaround is to do `make clean all`.~ Solved (fixed), thank's @keithf4 

There are (in the beginning of the branch) two commits with improvements of `pg_partman_bgw.c`. If you want to move them to another PR - let me know.

Pointing out to typos, comments, tips, etc. are welcome.


keithf4:
Thanks for this work. I'd been trying to figure out something to make the handling of the update files and individual files like this for a while. This is an interesting idea. Only thing I'm not 100% on is that the Makefile would have to be updated every single release.

vitaly-burovoy:
I made Makefile to looks like this to only be sure (and prove) changes from "updates/pg_partman--4.7.3--5.0.0-beta.sql" are correctly applied to individual files correctly.

Of course if this method will be accepted Makefile for future development process can be [slightly] different.

keithf4:
Sure I understand. Just giving my initial impression and some feedback. I definitely appreciate the work and look forward to having something to automate this process better. My method is a manual one using Meld, but it's worked for a long while now :)

keithf4:
So I did not mean to close this PR. It apparently auto-closed when I got rid of the 5.0 beta branch. If you'd like to do this PR against the 5.x release now that it's out, I'd be glad to revisit this process to help automate updates a little better.

PR #545, opened 2023-07-05
Fixes for some bugs [in 5.0.0-beta] for which solutions could be found quickly

vitaly-burovoy:
Though some fixes depend on project knowledge, e.g. 

sql/procedures/undo_partition_proc.sql:63
`v_target_schema`, `v_target_tablename` are set but never used but should (in branch when `p_target_table` is not set)

sql/functions/run_maintenance.sql:54
`v_sub_refresh_done` never assigned and used as NULL at row 361; 
row 367 is useless because `array_append` does not append inplace, it returns new array.

etc.

The branch is based on previous PoC #544, but still can be rebased on the "5.0.0-beta" branch if get changes from the file "updates/pg_partman--4.7.3--5.0.0-beta.sql" only. Two commits from there are required to reduce whitespace conflicts:
* 5ad4bdc Remove trailing spaces in actual SQL files
* 6d1038e Adjust file name with function name: dump_partitioned_table_definition


keithf4:
At this point, I'm not likely going to have the time to evaluate the Makefile changes for the initial 5.0 release. Could you please redo this PR (or make another one) to only be changing the update file? I have my own other method for getting those changes to the individual files for now.

vitaly-burovoy:
OK, I'd like to open a new PR (with new branch) and support both of them because if I find something strange I'll change this one and reflect changes to another one.

It greatly increases my performance, because I do not have your method to spread changes from update file to individual ones. 😇

One question: if changes from this branch can be applied to individual files without conflicts, may I leave them applied or I should create a branch with only "updates/pg_partman--4.7.3--5.0.0-beta.sql" changed?

keithf4:
> OK, I'd like to open a new PR (with new branch) and support both of them because if I find something strange I'll change this one and reflect changes to another one.
> 
> It greatly increases my performance, because I do not have your method to spread changes from update file to individual ones. innocent
> 
> One question: if changes from this branch can be applied to individual files without conflicts, may I leave them applied or I should create a branch with only "updates/pg_partman--4.7.3--5.0.0-beta.sql" changed?

Please do not touch any individual files in the PR for now. Thank you!

vitaly-burovoy:
I noticed the `partition_gap_fill` function is NOT declared (since 528116e) to be created in @extschema@ (schema name is missed), but it is created there (it seems Postgres adds SCHEMA name on `CREATE EXTENSION` to the `search_path` for this case).

Is it worth to add a commit touched "base" file to make declaration look similar to ones of other functions?

vitaly-burovoy:
This PR is still marked as `requires submitter response`. What response is expected from me?

keithf4:
Sorry, that was my mistake. I'll hopefully get more time to look at this next week.

vitaly-burovoy:
One more question... I'd like to submit some semantic improvements not related to fixing bugs but which increases readability like use
```sql
IF NOT FOUND THEN
```
instead of
```sql
IF v_jobmon IS NULL THEN
```
after selecting `jobmon` (to  `v_jobmon`) from table where it is NOT NULL.
Result will be same but the current code confuses because it "says" do something depending on something related to jobmon.

I guess these changes lead to increase size of "updates/pg_partman--4.7.3--5.0.0-beta.sql" because full function bodies will be there even if a several lines are changed and will be hard to understand why they are there.

Is it OK to send this changes here and then decide what to do with them (will be visible changes in sources) or I should create a new branch/PR from some commit not connected with fixing bugs, e.g. b43e488?

keithf4:
> One more question... I'd like to submit some semantic improvements not related to fixing bugs but which increases readability like use
> 
> ```sql
> IF NOT FOUND THEN
> ```
> 
> instead of
> 
> ```sql
> IF v_jobmon IS NULL THEN
> ```
> 
> after selecting `jobmon` (to `v_jobmon`) from table where it is NOT NULL. Result will be same but the current code confuses because it "says" do something depending on something related to jobmon.
> 
> I guess these changes lead to increase size of "updates/pg_partman--4.7.3--5.0.0-beta.sql" because full function bodies will be there even if a several lines are changed and will be hard to understand why they are there.
> 
> Is it OK to send this changes here and then decide what to do with them (will be visible changes in sources) or I should create a new branch/PR from some commit not connected with fixing bugs, e.g. [b43e488](https://github.com/pgpartman/pg_partman/commit/b43e4889ed1a7da3346a34a98a06d94e2b22e367)?

If you could wait for any changes like this that won't be part of 5.0 until 5.0 has been released, I'd appreciate it. Trying to include it with this would just be a bit confusing.

Not sure I like that change to NOT FOUND personally. Doesn't really give any benefit that I'm aware of and I think is less clear about what is being used to make the conditional decision. I think the NOT FOUND conditional is better for simpler, one-off query result evaluation, not when it's part of larger plpgsql functions.

vitaly-burovoy:
Oops. It seems I get wrong words... 😇
Of course I will create a new PR.

I mean whether it is OK to push on top of this branch (I will keep it updated) to decrease possible conflicts or create a new branch from a commit below commits with fixes?

P.S.: the query I sent example is quite small:
https://github.com/pgpartman/pg_partman/blob/8430165fdf2309932318dd943d79f73e619dbcdc/sql/functions/apply_privileges.sql#L30
```sql
SELECT jobmon INTO v_jobmon FROM @extschema@.part_config WHERE parent_table = p_parent_schema ||'.'|| p_parent_tablename;
IF v_jobmon IS NULL THEN
    RAISE EXCEPTION 'Given table is not managed by this extention: %.%', p_parent_schema, p_parent_tablename;
END IF;
```
which actually looks like this: "if nothing was about 'jobmon', then `table is not managed by this extention`"...

keithf4:
This seems more a personal preference for the code than any real technical benefit (unless I'm wrong which please clarify then). It just seems clearer what's going on, especially when you don't come back to look at code for long periods of time. And it keeps it consistent with how it is done elsewhere.

Please just wait on any style/semantic changes like this until after 5.0.0 is released. Don't get me wrong, I appreciate your contribution, but this is a big change already so I'd prefer stylistic code updates that don't change core functionality to wait until the other bigger work is done. They'll be much easier to review.

vitaly-burovoy:
OK, I'll wait the release. Style changes can certainly wait.

PR #546, opened 2023-07-05
Fixes for some bugs [in 5.0.0-beta] for which solutions could be found quickly [port from #545]

vitaly-burovoy:
Though some fixes depend on project knowledge, e.g.

sql/procedures/undo_partition_proc.sql:63
v_target_schema, v_target_tablename are set but never used but should (in branch when p_target_table is not set)

sql/functions/run_maintenance.sql:55
v_sub_refresh_done never assigned and used as NULL at row 376;
row 382 is useless because array_append does not append inplace, it returns new array.

etc.

This branch is a reflection of work done in #545 without commits made in (and changes based on) #544.

Upd.: adjusted line numbers pointing to original files has not touched by #545


vitaly-burovoy:
Sorry, initially forgot to include changes in "Makefile" and in "pg_partman_bgw.c" which are useful/important and not related to migration generator.

Force pushed with them.

Makefile
```diff
@@ -15,20 +19,16 @@ ifneq ($(NO_BGW),)
...
 
+.PHONY: all
 all: sql/$(EXTENSION)--$(EXTVERSION).sql
 
 SCRIPTS = bin/common/*.py
 
-sql/$(EXTENSION)--$(EXTVERSION).sql: $(sort $(wildcard sql/types/*.sql)) $(sort $(wildcard sql/tables/*.sql)) $(sort $(wildcard sql/functions/*.sql)) $(sort $(wildcard sql/procedures/*.sql))
-	cat $^ > $@
+sql/$(EXTENSION)--$(EXTVERSION).sql: | $(wildcard sql/types/*.sql sql/tables/*.sql sql/functions/*.sql sql/procedures/*.sql)
+	cat $| > $@
```
keithf4, line 28: So I specifically had the sort operations on each of the individual types of files because I wanted a predictable way to have certain operations be first for each type of object.  Since extension updates/installs are all done as one giant transaction, this was the only way I could have to be able to guarantee a specific point that certain operations could run. I could put those operations in the alphabetically first file of that type and it would guarantee to be first.

keithf4 (changes requested):

Makefile
keithf4: This was a necessity early on in some development. Don't think I've had to worry about this for a while, but I would like to keep the functionality there if needed

keithf4: Testing it out, it looks like the `$|` does the same thing. But just want to make sure the end result always the same as what the previous options did.

vitaly-burovoy: The docs says[1] the returned list is sorted (the link is mentioned in the commit[2]):

> As with wildcard expansion in rules, the results of the wildcard function are sorted. But again, each individual expression is sorted separately, so ‘$(wildcard *.c *.h)’ will expand to all files matching ‘.c’, sorted, followed by all files matching ‘.h’, sorted. 

---
[1] https://www.gnu.org/software/make/manual/html_node/Wildcard-Function.html
[2] https://github.com/pgpartman/pg_partman/commit/525d92a51618f32b31b62a99fa73b122eb75ebb8

vitaly-burovoy: > are never checked

Ah! I missed that sentence. OK, we can leave "$^", but still use a single "$(wildcard …)" call… 😇

vitaly-burovoy:
Moved back "$|", left deletion redundant "$(sort …)" and multiple "$(wildcard …".

Makefile
keithf4: Please go back and re-read my initial reasoning for the sorting being there. I would like to keep the functionality to allow a predictable method of running code at a specific point during an update or install.

vitaly-burovoy: Hmm. If I understand docs correctly, there is nothing changed in sorting order.

The only reason why I tried to use "$|" was an attempt to [possible] guard against multijob (`make -j5`). But if `make` will sort prerequisites by the time when they are ready without a pipe symbol it will do it for prerequisites with your multiple `$(sort …)` entries because they are just gives a list of required targets which will be the same as when using a single `$(wildcard …)` and multiple rules inside it.

Please could you show how my change leads to a different result.

vitaly-burovoy: P.S.: I'd like to use the single "$(wildcard …)" because its behavior is the same as you wrote in the initial reasoning.

I posted an excerpt from docs about expected behavior in this discussion above:
https://github.com/pgpartman/pg_partman/pull/546#discussion_r1253295553

keithf4: Hmm ok I missed that that was part of the wildcard doc itself and not the `$|` entry.

Are you sure that's consistent for all make or just gnu make? I could've sworn the reason I did that was because I was just getting arbitrary file name lists back with no predictable ordering.

keithf4: Ok. I doubt PG is making use of the actual order that things are put together in any significant way. Haven't had to worry about the order for a long time, so I guess if it ever comes up again in the future I'll explore it again

updates/pg_partman--4.7.3--5.0.0-beta.sql
```diff
@@ -3549,23 +3580,16 @@ IF p_batch_interval IS NULL OR p_batch_interval > v_partition_interval THEN
...
+SELECT string_agg(quote_ident(attname), ',')
+INTO v_column_list
+FROM pg_catalog.pg_attribute a
+JOIN pg_catalog.pg_class c ON a.attrelid = c.oid
+JOIN pg_catalog.pg_namespace n ON c.relnamespace = n.oid
+WHERE n.nspname = v_source_schemaname
+AND c.relname = v_source_tablename
+AND a.attnum > 0
+AND a.attisdropped = false
+AND attname <> ALL(COALESCE(p_ignored_columns, ARRAY[]::text[]));
```
keithf4, line 3592: Fantastic simplification here. Thank you!

updates/pg_partman--4.7.3--5.0.0-beta.sql
```diff
@@ -4011,10 +4024,12 @@ $$;
 
 CREATE FUNCTION @extschema@.run_maintenance(
     p_parent_table text DEFAULT NULL
+    -- If these default chaged reflect them in `run_maintenance_proc`!
```
keithf4, line 4027:
```suggestion
    -- If these default changed reflect them in `run_maintenance_proc`!
```

updates/pg_partman--4.7.3--5.0.0-beta.sql
```diff
@@ -5462,6 +5452,54 @@ END;
...
+
+BEGIN
+
+v_adv_lock := pg_try_advisory_lock(hashtext('pg_partman run_maintenance'));
+IF v_adv_lock = false THEN
+    RAISE NOTICE 'Partman maintenance already running or another session has not released its advisory lock.';
+    RETURN;
+END IF;
+
+FOR p_parent_table IN
```
keithf4, line 5476:
```suggestion
FOR v_parent_table IN
```
Please use `v_` prefix for variables that aren't parameters

updates/pg_partman--4.7.3--5.0.0-beta.sql
```diff
@@ -5462,6 +5452,54 @@ END;
...
+LOOP
+/*
+ * Run maintenance with a commit between each partition set
+ * TODO - Once PG11 is more mainstream, see about more full conversion of run_maintenance function as well as turning
+ *        create_partition* functions into procedures to commit after every child table is made. May need to wait
+ *        for more PROCEDURE features as well (return values, search_path, etc).
+ *      - Also see about swapping names so this is the main object to call for maintenance instead of a function.
+ */
+    RAISE DEBUG 'run_maintenance_proc for table: %', p_parent_table;
+    PERFORM @extschema@.run_maintenance(p_parent_table, p_jobmon => p_jobmon, p_analyze => p_analyze);
```
keithf4, line 5490: So I kind of wanted the actual, full call of the function to be part of the DEBUG statement here. So I know all the parameters that were passed as part of the call. Makes it a bit more complicated to create the SQL but better for debugging.

keithf4 (changes requested):
Thanks so much for such an extensive review of the update code! Really got rid of a lot of cruft and cleaned up formatting nicely. Will continuing reviewing your comments in the PR message when I have a chance.

updates/pg_partman--4.7.3--5.0.0-beta.sql
vitaly-burovoy: Oops. Will fix it.
=)

updates/pg_partman--4.7.3--5.0.0-beta.sql
```diff
@@ -5462,6 +5452,54 @@ END;
...
+    -- Keep these defaults in sync with `run_maintenance`!
+    , p_analyze boolean DEFAULT false
+    , p_jobmon boolean DEFAULT true
+)
+    LANGUAGE plpgsql
+    AS $$
+DECLARE
+
+v_adv_lock              boolean;
+p_parent_table          text;
```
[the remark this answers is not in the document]
vitaly-burovoy: Looks like a copy-paste error. Will fix it.

vitaly-burovoy:
Thank you for many warm words!
Are you comfortable if I push-force changes or should I wait until you finish reviewing the PR?

src/pg_partman_bgw.c
```diff
@@ -425,7 +425,7 @@ void pg_partman_bgw_run_maint(Datum arg) {
             , dbname);
 
     resetStringInfo(&buf);
-    appendStringInfo(&buf, "SELECT n.nspname FROM pg_catalog.pg_extension e JOIN pg_catalog.pg_namespace n ON e.extnamespace = n.oid WHERE extname = 'pg_partman'");
+    appendStringInfo(&buf, "SELECT pg_catalog.quote_ident(nspname) FROM pg_catalog.pg_extension e JOIN pg_catalog.pg_namespace n ON e.extnamespace = n.oid WHERE extname = 'pg_partman'");
```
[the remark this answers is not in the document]
keithf4: That they're not being used is a lack of consistency on my part. As you can tell from the cleanup you've been doing, it's not being the greatest.

Aliases have nothing to do with the schema. It's to avoid ambiguity amongst column names being similar when joining tables. None are now, but it avoids any possibility of it ever being an issue in the future.

updates/pg_partman--4.7.3--5.0.0-beta.sql
keithf4: If the problem happens back in the actual caller of the function due to some bad value being passed in, then the call stack will never reach the called function to be debugged. And call tracing is particularly bad in plgsql because you'll get such lovely simplistic errors as `ERROR at .` with no further reference to line number. You do get a character count number in the log sometimes but even that can often be several lines off where the actual issue is.

I don't get the dislike of building SQL in order to debug. I know it should be done carefully with format. Also I know DEBUG is usually disabled, but the entire reason it's there is so that it can be enabled when it is needed. If debugging were better in plgsql I could see your point, but when building more complicated SQL statements, having that SQL output during debug parsing can be greatly useful and I often have DEBUG enabled during development for that purpose.This particular instance isn't that complicated, so just debugging the parameters should probably be ok. But I don't think the practice should be summarily dismissed.

keithf4:
> Thank you for many warm words! Are you comfortable if I push-force changes or should I wait until you finish reviewing the current state of the PR?

Please don't force push into development into development. You can do the changes to this PR and once it's approved then it can be merged.

I should probably fix this repo to require at least one approved review before merging a PR and not allow direct pushes. Just haven't had much development outside myself :)

keithf4:
Looking into these...

> Though some fixes depend on project knowledge, e.g.
> 
> sql/procedures/undo_partition_proc.sql:63 v_target_schema, v_target_tablename are set but never used but should (in branch when p_target_table is not set)
> 

So `v_target_schema` is never used, but `v_target_tablename` is used to determine whether the given object exists and raising an exception. I've removed the schema variable since it's not used.

> sql/functions/run_maintenance.sql:55 v_sub_refresh_done never assigned and used as NULL at row 376; row 382 is useless because array_append does not append inplace, it returns new array.

`v_sub_refresh_done` is an array variable and has values appended to it. It used to avoid needlessly refreshing a subscription multiple times if several partition sets are part of the same subscription. Unless you see that it's not happening somehow?

keithf4:
> 
> sql/functions/run_maintenance.sql:55 v_sub_refresh_done never assigned and used as NULL at row 376; row 382 is useless because array_append does not append inplace, it returns new array.

Good catch. I've replaced it with
```
v_sub_refresh_done := array_append(v_sub_refresh_done, v_row.subscription_refresh);
```

updates/pg_partman--4.7.3--5.0.0-beta.sql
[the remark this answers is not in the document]
keithf4: Just checking in to see if you'll be able to apply the requested changes for this PR

src/pg_partman_bgw.c
keithf4, line 428:
```suggestion
    appendStringInfo(&buf, "SELECT pg_catalog.quote_ident(n.nspname) FROM pg_catalog.pg_extension e JOIN pg_catalog.pg_namespace n ON e.extnamespace = n.oid WHERE e.extname = 'pg_partman'");
```

updates/pg_partman--4.7.3--5.0.0-beta.sql
keithf4: Will be reverting this debug change after merging

keithf4 (changes requested):

keithf4:
Thanks again for all the work you put into this!

vitaly-burovoy:
When you merged this PR onto f4f5517fe9fc8a69c03c02865c5f8b6a9ffca484, quotes around "%s" for inserting schema name were not removed (`quote_ident` adds it when necessary) but a new line was added instead. Please fix it this way (I do not think it worth another PR):
```patch
diff --git a/src/pg_partman_bgw.c b/src/pg_partman_bgw.c
index 693d322..a17de21 100644
--- a/src/pg_partman_bgw.c
+++ b/src/pg_partman_bgw.c
@@ -490,2 +490 @@ void pg_partman_bgw_run_maint(Datum arg) {
-
-    appendStringInfo(&buf, "CALL \"%s\".run_maintenance_proc(p_wait => %d, p_analyze => %s, p_jobmon => %s)", partman_schema, pg_partman_bgw_maintenance_wait, analyze, jobmon);
+    appendStringInfo(&buf, "CALL %s.run_maintenance_proc(p_wait => %d, p_analyze => %s, p_jobmon => %s)", partman_schema, pg_partman_bgw_maintenance_wait, analyze, jobmon);
```

---
Thank you too for reviewing!

P.S.: what does the label "in development" mean (which is still applied)? I was waiting for the "requires submitter response" to appear… 😇

keithf4:
> When you merged this PR onto [f4f5517](https://github.com/pgpartman/pg_partman/commit/f4f5517fe9fc8a69c03c02865c5f8b6a9ffca484), quotes around "%s" for inserting schema name were not removed (`quote_ident` adds it when necessary) but a new line was added instead. Please fix it this way (I do not think it worth another PR):
> 
> ```diff
> diff --git a/src/pg_partman_bgw.c b/src/pg_partman_bgw.c
> index 693d322..a17de21 100644
> --- a/src/pg_partman_bgw.c
> +++ b/src/pg_partman_bgw.c
> @@ -490,2 +490 @@ void pg_partman_bgw_run_maint(Datum arg) {
> -
> -    appendStringInfo(&buf, "CALL \"%s\".run_maintenance_proc(p_wait => %d, p_analyze => %s, p_jobmon => %s)", partman_schema, pg_partman_bgw_maintenance_wait, analyze, jobmon);
> +    appendStringInfo(&buf, "CALL %s.run_maintenance_proc(p_wait => %d, p_analyze => %s, p_jobmon => %s)", partman_schema, pg_partman_bgw_maintenance_wait, analyze, jobmon);
> ```
> 
> Thank you too for reviewing!
> 
> P.S.: what does the label "in development" mean (which is still applied)? I was waiting for the "requires submitter response" to appear… 😇

So, glad I actually tested this. I installed partman to the `PartMan` schema and adjusted the bgw id test to set the schema appropriately. Had to fix some schema qualified calls in it as well (see commit https://github.com/pgpartman/pg_partman/commit/820f2c770daf5006a306725570980c60f0ab69d5)
```
SELECT set_config('search_path','"PartMan", public',false);
```
So, yeah it breaks as expected without removing the quotes if quote_ident is used. However it also does not work when the quotes are removed. The only way I can currently get it working is without using quote_ident() and escaping the quotes as was done before. Are you able to find a working method using quote_ident()?

with quote_ident():
```
2023-08-21 12:30:29.896 EDT [2023-08-21 12:30:29 EDT]  [1150341] user=,db=,app=,err=42601,sess=64e39125.118d85: ERROR:  syntax error at or near "" at character 6
2023-08-21 12:30:29.896 EDT [2023-08-21 12:30:29 EDT]  [1150341] user=,db=,app=,err=42601,sess=64e39125.118d85: QUERY:  CALL "PartMan"~!D.run_maintenance_proc(p_wait => 5, p_analyze => true, p_jobmon => true)
```

Just escaping double-quotes:
```
2023-08-21 12:30:52.103 EDT [2023-08-21 12:30:52 EDT]  [1150472] user=,db=,app=,err=00000,sess=64e3913c.118e08: LOG:  pg_partman dynamic background worker (dbname=keith): CALL "PartMan".run_maintenance_proc(p_wait => 5, p_analyze => true, p_jobmon => true) called by role keith on database keith
```

vitaly-burovoy:
> QUERY:  CALL "PartMan"~!D.run_maintenance_proc

I guess the error is not in the `quote_ident` itself, but in using `DatumGetCString` instead of `TextDatumGetCString` in line 465.

The first macros returns a pointer to a stored value "as is" and can be used if returned type is `pg_catalog.cstring`, but if there is not a null-terminated string (like TEXT, NAME, etc.), sometimes zero bytes is there , but sometimes not. The second macros copies data and set zero-terminated byte after it.

As I can see garbage chars are inserted after correctly quoted schema, I faced similar behavior (garbage chars with incorrect macros) years ago.

Can test in several hours, not now.

keithf4:
Now I'm really confused how I even got this working before. With or without your change, if you actually have a partition set to run maintenance on, it's throwing this error when partman is installed to the `PartMan` schema. 
```
2023-08-22 16:06:04.139 EDT [2023-08-22 16:06:04 EDT]  [1246475] user=,db=,app=,err=3F000,sess=64e5152c.13050b: ERROR:  schema ""PartMan"" does not exist at character 8
2023-08-22 16:06:04.139 EDT [2023-08-22 16:06:04 EDT]  [1246475] user=,db=,app=,err=3F000,sess=64e5152c.13050b: QUERY:  SELECT """PartMan""".run_maintenance('partman_test.id_taptest_table', p_jobmon := 't', p_analyze := 't')
```
It's not just the background worker. Manual CALLs of `run_maintenance_proc()` also fail as well. 
```
keith=# call "PartMan".run_maintenance_proc();
ERROR:  schema ""PartMan"" does not exist
LINE 1: SELECT """PartMan""".run_maintenance('partman_test.id_taptes...
               ^
QUERY:  SELECT """PartMan""".run_maintenance('partman_test.id_taptest_table', p_jobmon := 't', p_analyze := 'f')
CONTEXT:  PL/pgSQL function "PartMan".run_maintenance_proc(integer,boolean,boolean) line 38 at EXECUTE
```
This is because the creation of the extension requires the command
```
create extension pg_partman schema "PartMan";
```
Which the makes all the internal references to `@extschema@` turn into `"PartMan"`. That then breaks the call in `run_maintenance_proc()` because it's doing this
```
    v_sql := format('SELECT %I.run_maintenance(%L, p_jobmon := %L',
        '"PartMan"', v_row.parent_table, p_jobmon);
```
Does this back on version 4.7.3 as well, so it's not any changes in 5.0.0-beta.

keithf4:
So I went through and looked at how many places `@extschema@` was actually being used with `%I` or `quote_ident`. Only 4 for the first and none for the latter. Since it seems to do a direct string replacement in all case, including double-quotes, I just used the `%s` macro instead for now. Can see about maybe doing it more securely by trimming quotes and going back to `%I` some other time, but the inherent quoted replacement of it in so few places doesn't have me worried at this point.

Added  a new test for if partman is installed in a mixed-case schema while using the BGW if you'd like to try it. Note at the top has how partman has to be installed to use it.

vitaly-burovoy:
Ough. Also my fault, I just checked background worker correctly calls `run_maintenance_proc`, but `part_config` was empty and I did not dig deeper.

It seems "@extschema@" is already quoted (when used for substitution), because it is correctly used (and intended¹ to use) in regular (not string formatted) queries as schema name, e.g.
https://github.com/pgpartman/pg_partman/blob/eee898a2b5ed6c80d73a8fb171d91c9dd8c3e9f8/updates/pg_partman--4.7.3--5.0.0-beta.sql#L767

So "@extschema@" does not need to be specially formatted and can be inserted directly to DDL strings² instead of formatting parameter, like in
https://github.com/pgpartman/pg_partman/blob/eee898a2b5ed6c80d73a8fb171d91c9dd8c3e9f8/updates/pg_partman--4.7.3--5.0.0-beta.sql#L1790

---
> Can see about maybe doing it more securely by trimming quotes

Since 2023-08-07 it is not required anymore because now characters which could lead to vulnerabilities are forbidden (backported to all supported versions since v11):
https://git.postgresql.org/gitweb/postgresql?p=postgresql.git;a=commit;h=cd5f2a357014d5ed8205191f8e1fb180f1439599

---
> So I went through and looked at how many places

It is hard for me to do so with the current state of the project where "sql/\*/\*.sql" files are partially obsolete and new version of functions are only in "updates/….sql". Hope #544 will be merged soon (I will adjust it after 5.0.0 is released).

---
¹ See usage of "@extschema@" in SET without quotes:
https://www.postgresql.org/docs/15/extend-extensions.html#EXTEND-EXTENSIONS-RELOCATION

² I still think it is bad practice to format query string before call. I got you point about debugging (and call stack) and will think how to do it better.

keithf4:
Thanks for the reference for that update that made the macro usage more secure! And also thanks again for finding the proper C function to use with quote_ident before as well. I'm still lacking in that area of PG internals knowledge.

To be honest, I'm probably not going to be merging #544 as it is yet. I do like the automation work you did and will definitely look at adapting it for future releases. But for the 5.0.0 release I'd like to use my old method for now just to be able to get it done quicker (for me). Can re-evaluate that PR once 5.0.0 is out. Thanks!

vitaly-burovoy:
> Thanks

My pleasure! 😁

---
> To be honest, I'm probably not going to be merging #544 as it is yet.

Of course! I meant "soon after 5.0.0 is released". 😇

Anyway, my version or any other automation is highly desirable!

PR #547, opened 2023-07-19
Readme Suggestions

andyatkinson:
- Typo
- Link the doc folder in markdown
- Link markdown
- Add more Markdown links

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,16 @@ Recommended:
 
 In the directory where you downloaded pg_partman, run
 
-    make install
+```sh
+make install
+```
 
 If you do not want the background worker compiled and just want the plain PL/PGSQL functions, you can run this instead:
 
-    make NO_BGW=1 install
+
+```sh
+make NO_BGW=1 install
+```
 
 The background worker must be loaded on database start by adding the library to shared_preload_libraries in postgresql.conf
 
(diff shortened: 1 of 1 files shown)


andyatkinson:
Hi @keithf4 👋 - I finally started to look at the pg_partman project although I haven't yet put it into action to partition a table, but hav plans to. My goal is to provide you some user feedback and offer suggestions where I can as a way to say "thank you" for the open source work you and the contributors and maintainers do here!

Besides these commits above that fix a typo, and make some stylistic choices with more Markdown formatting and syntax highlighting, I wanted to provide some feedback on the README based on my initial reactions.


## On Preferring "Declarative Partitioning" over "Native"

For the use of “Native” had you considered preferring the term "Declarative Partitioning"? The PostgreSQL docs use this term. When I was learning about partitioning, “Native” didn't carry a lot of meaning to me because I didn't know what the distinction was all about. [PostgreSQL: Documentation: 15: 5.11. Table Partitioning](https://www.postgresql.org/docs/current/ddl-partitioning.html)

The partitioning methods that existed before Declarative Partitioning was build in, e.g. trigger functions, table inheritance, etc. are very unlikely to be techniques new users would reach for now. Do you agree? I think for new learners they are unlikely to ever use the non-built in method.

Plus I think using the same terminology the Postgres docs use helps reinforce the learning. Having a Guide on "Legacy Methods" or whatever that discuss trigger-based and explain the native/non-native distinction in the doc folder makes sense to me, but perhaps de-emphasized from the top level Readme. Just a thought!


## **UPGRADE** Section

Upgrades wouldn't be relevant for new users looking to install partman for the first time. They'd want to know what the current version is and begin using it. I'd suggest deemphasizing this unless most of the users are partman users that are upgrading from old versions. I don't have that context. 

The Upgrade guides are very valuable though for existing users. I'd consider turning this into a table (with Markdown) for easier reading. The table could show the current version, and then there could be linked upgrade guides from past source and target versions. This feels more like a section for the “doc” folder as a guide that's linked to from the top-level Readme.

## TESTING Section

> This extension can use the pgTAP unit testing suite to evaluate if it is working properly ( [http://www.pgtap.org](http://www.pgtap.org/) ).

This is cool but could you expand on this and show readers how to do that with a guide? I know that pgtap unit testing exists but I don't know how to use it to evaluate whether pg_partman is working properly. Perhaps this could be a Guide in the doc folder as well.


Just some initial thoughts. If that was helpful let me know and I can send more ideas! Take these with a grain of salt and as first time Readme user feedback.

Thanks!

keithf4:
Appreciate the feedback and the PR!

The next version of pg_partman will be dropping trigger-based partitioning all together, so there will no longer be any references to it anywhere other than the migration guides. As for renaming "native" to "declarative", I can see about doing that after 5.0 is out since as things are now, I use the term "native" a lot within the application itself.

Upgrading is a huge part of maintaining extensions, so I don't know that I'll change it too much on the main README. But maybe check how I've done things in the 5.0 release files?  I could see about adding a table to explain some of the docs better https://github.com/pgpartman/pg_partman/tree/5.0.0-beta

Expanding on the testing would be good. I'll make an issue so I don't forget to try and come back to do that

keithf4 (approved):

ChuckHend:
Nice work @andyatkinson 🚀

PR #560, opened 2023-09-01
optimize_constraints starts with 0

jw1u1:
make clear that optimize_constraints has to be set to zero to process the last partition.


keithf4:
So I do have a clearer explanation of this, but it's down in the `apply_constraints()` section of the docs. Could move that up to here and I think that would be a better explanation. I'd want to be cautious on mentioning using a `0` value since setting it to that would apply the constraint on the current table and potentially block out new data. I may even consider making a constraint to disallow this setting.

jw1u1:
0 is wrong, guessed that due to param +1 in the code...

PR #561, opened 2023-09-02
Use [in docs] "format" instead of concatenation for relation names

vitaly-burovoy:
Also remove [from docs] unnecessary conversion "oid" columns to "regclass" type when searching in tables: regclass is converted back to the "oid" type for comparison (and to use indexes).

The best way to see changes is:
git diff --word-diff=color HEAD^..HEAD

### DIFF
--- a/doc/migrate_to_declarative.md
+++ b/doc/migrate_to_declarative.md
@@ -93,7 +93,7 @@ The first major step in this migration process is now to uninherit all the child
 DO NOT RUN THE RESULTING STATEMENTS YET. A future query will not work if the child tables are no longer part of the inheritance set.
 
 ```
-SELECT 'ALTER TABLE '||inhrelid::regclass||' NO INHERIT '||inhparent::regclass||';' 
+SELECT format('ALTER TABLE %s NO INHERIT %s;', inhrelid::regclass, inhparent::regclass)
 FROM pg_inherits 
 WHERE inhparent::regclass = 'partman_test.time_taptest_table'::regclass;
                                               ?column?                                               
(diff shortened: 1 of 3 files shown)


vitaly-burovoy:
One question for changed queries: I left DDL generation logic "AS IS", but do not understand one block:
```sql
to_char(to_timestamp(substring(c.relname from 10), 'YYYYMMDD'), 'YYYYMMDD')
```
https://github.com/pgpartman/pg_partman/blob/3427224b407a094a36c1a4a084d6a1e0ac407a78/doc/migrate_to_partman.md?plain=1#L134

It looks as no-op for me (could be replaced with just `substring(c.relname from 10)`).

Whether it is copy-paste error or it is intentional and has special behavior?

keithf4:
> One question for changed queries: I left DDL generation logic "AS IS", but do not understand one block:
> 
> ```sql
> to_char(to_timestamp(substring(c.relname from 10), 'YYYYMMDD'), 'YYYYMMDD')
> ```
> 
> https://github.com/pgpartman/pg_partman/blob/3427224b407a094a36c1a4a084d6a1e0ac407a78/doc/migrate_to_partman.md?plain=1#L134
> 
> It looks as no-op for me (could be replaced with just `substring(c.relname from 10)`).
> 
> Whether it is copy-paste error or it is intentional and has special behavior?

So this was not a good example here since, yes, this would essentially be a no-op because both the old time pattern and the new one are the same. If instead the time pattern was `YYYY_MM_DD` or any other pattern that someone may have been using, you'd want to do the full conversion steps to ensure the new time suffix matches the actual time value of what the old one was. I'll try and update this with a better example that's not a no-op. Thanks!

keithf4:
Pushed up a better example for this so it doesn't look like a no-op. Thanks again!
Also went and added some pre-commit hooks to automatically clean up whitespace.

vitaly-burovoy:
Thank you!

PR #589, opened 2023-11-12
Allow overriding of pg_config

ChuckHend:
Allow user to override the pg_config using in Makefile. For example:

```bash
PG_CONFIG=/path/to/my/pg_config make install
```

@keithf4 , maybe there's a way to do this already? Some other C based Postgres extensions allow this, e.g. [pg_cron](https://github.com/citusdata/pg_cron/blob/a8cb0e0b5018cf82e5a90728a91d04fb79594642/Makefile#L22) and [pgvector](https://github.com/pgvector/pgvector/blob/69a2ce0d432f45f24186ce7ef1683ab9af5329df/Makefile#L46)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ EXTENSION = pg_partman
 EXTVERSION = $(shell grep default_version $(EXTENSION).control | \
                sed -e "s/default_version[[:space:]]*=[[:space:]]*'\([^']*\)'/\1/")
 
-PG_CONFIG = pg_config
+PG_CONFIG ?= pg_config
 PG_VER = $(shell $(PG_CONFIG) --version | sed "s/^[^ ]* \([0-9]*\).*$$/\1/" 2>/dev/null)
 
 PG_VER_min = 14


andyatkinson:
@ChuckHend 👋 Nice. Seems useful. Is there a security concern with that?

Do you have a fork of pg_partman or example app where you're showing how you're using the overridden pg_config values (or do those projects offer that?)? Mostly I'm curious about what the use case is, but it might also server as documentation or changelog documentation for @keithf4 / for the project, if accepted.

ChuckHend:
> Do you have a fork of pg_partman or example app where you're showing how you're using the overridden pg_config values (or do those linked projects offer that?)?

Here's an example where we change the value in the Makefile directly, which is messy https://github.com/tembo-io/pgmq/blob/293f6e93f3799ee17016b07f4834f7bd01f7387a/.github/workflows/extension_ci.yml#L94

This is what we'd like to do:

```
git clone pg_partman
cd pg_partman
make install PG_CONFIG=/path/to/my/specific/pg_config
```

Instead of having PG_CONFIG be the config that is on PATH.

That's easy to do with pg_cron and pg vector, because their Makefiles are set up the way I am proposing in this PR.

keithf4:
This seems like a reasonable request. Thanks!

keithf4 (approved):

keithf4:
Approved. Will include this update in the next release. Don't have an ETA for it at this time.

Issue #519, opened 2023-04-03
partition_data_proc and run_maintenance with infinity timestamps


jw1u1:
partition_data_proc and run_maintenance seems not to handle infinity timestamps.
While 
bre=# call partman.partition_data_proc('foo');
ERROR:  Attempted partition time interval is outside PostgreSQL's supported time range. 
                        Unable to create partition with interval before timestamp 294277-01-01 00:00:00+01 

infinite_time_partitions = false does't help.

Expectation: infinity timestamps should remain in the default partition and new partitions should be created up to max(timestamp) where timestamp != 'infinity'.

keithf4:
The default partition feature is built into postgresql and the constraint on the default is always the anti-constraint of all other child tables combined. So that will not work to try and assign that boundary to the built in default.

I'm honestly not sure if PostgreSQL itself can support a boundary of infinity. Would have to test.

jw1u1:
> Please see the documentation for what `infinite_time_partitions` is for. It doesn't have anything to do with the infinity value. It's to tell pg_partman to keep making time-based partitions infinitely even if there is no new data.

agree, but could prevent to try to create infinity child tables.

keithf4:
Yes, the value itself is a supported thing in PostgreSQL. But does that table have any constraint boundaries on the `event_ts` column?

keithf4:
So looking into this more, I see how you're expecting this to work. However, as you saw, an infinite value cannot be partitioned out. By definition `infinity` is greater than all possible values, so how would it know which child table to put it into? This would only work on an unbounded child table, which is what the default table is. See example here:
```
keith=# \d+ mytable_default 
                                      Table "public.mytable_default"
  Column   |  Type   | Collation | Nullable | Default | Storage | Compression | Stats target | Description 
-----------+---------+-----------+----------+---------+---------+-------------+--------------+-------------
 dt        | date    |           | not null |         | plain   |             |              | 
 table_id  | integer |           |          |         | plain   |             |              | 
 date_from | bigint  |           |          |         | plain   |             |              | 
 date_to   | bigint  |           |          |         | plain   |             |              | 
 otec_id   | integer |           |          |         | plain   |             |              | 
 standard  | integer |           |          |         | plain   |             |              | 
 place     | integer |           |          |         | plain   |             |              | 
 count     | bigint  |           |          |         | plain   |             |              | 
Partition of: mytable DEFAULT
[...]

PR #593, opened 2023-11-29
create_parent: allow nullable control column if p_default_table is true

madand:
The current behavior of requiring a control column to be NOT NULL looks like a regression in v5.0.0.

### DIFF
--- a/sql/functions/create_parent.sql
+++ b/sql/functions/create_parent.sql
@@ -108,7 +108,7 @@ JOIN pg_catalog.pg_namespace n ON c.relnamespace = n.oid
 WHERE c.relname = v_parent_tablename::name
 AND n.nspname = v_parent_schema::name
 AND a.attname = p_control::name;
-    IF (v_notnull = false OR v_notnull IS NULL) THEN
+    IF (p_default_table = false AND (v_notnull = false OR v_notnull IS NULL)) THEN
         RAISE EXCEPTION 'Control column given (%) for parent table (%) does not exist or must be set to NOT NULL', p_control, p_parent_table;
     END IF;
 


keithf4:
Requiring the control column to be NOT NULL is not a regression. What is the use-case where having a default table would need to allow the control column to be NULL?

madand:
Our use case (that works ok with pg_pratman v4.x) is: we have an archive table that mirrors data from `pg_cron`'s table:
```
CREATE TABLE cron.job_run_details (
	jobid bigint,
	runid bigint primary key default pg_catalog.nextval('cron.runid_seq'),
	job_pid integer,
	database text,
	username text,
	command text,
	status text,
	return_message text,
	start_time timestamptz,
	end_time timestamptz
);
```
It is partitioned on `end_time`, which equals `NULL` in case `pg_cron` was unable to start the job at all.

Having a `DEFAULT` partition for handling these `NULL`s seems natural to me.  It would be great if `pg_partman` v5.x could support this.

keithf4:
Hmm... that would have been a bug if that was being allowed before. From experience I highly discourage keeping data that is actively used in the default table. It can cause serious performance issues if the data builds up to any significant degree. Every time a new child table is added, all data that exists in the default must be fully evaluated again to see if it matches the new child table's constraint. 

I've pretty much decided to enforce the NOT NULL state of things in partman when people have asked because I've also run into too many people running into the performance issues when they let  data build up in the default, NULL being one of the culprits of this. The default is there to catch accidental data that users don't want to lose but should be evaluated as to whether it is kept and moved to the proper child table or dropped. I'd recommend finding some alternative value to NULL, perhaps the minimal epoch? There could then be a dedicated partition for those values.
```
select to_timestamp(0);
      to_timestamp      
------------------------
 1969-12-31 19:00:00-05
```

keithf4:
I'm also looking to support the +/- infinitely values at some point too. Perhaps that could work  
https://github.com/pgpartman/pg_partman/issues/519

madand:
Thanks for the explanation, now I can see your motivation behind going the strict route.
We have decided to convert our control column to NOT NULL. So I'm closing the PR.

PR #607, opened 2024-01-09
Make table analyze controlled by parameter in partition_gab_fill func…

aleszeleny:
…tion.


keithf4:
I've implemented this in the upcoming 5.1 release. Thanks for providing the PR to work from!

Issue #387, opened 2021-09-20
Add support for CONCURRENT DETATCH in PG14


keithf4:
New syntax in PG14 can concurrently detach child tables. Make this the default for PG14 in pg_partman.

ALTER TABLE ... DETACH PARTITION ... CONCURRENTLY

keithf4:
Unfortunately it seems like this feature cannot be done via an extension function at this time.
```
keith@keith=# SELECT partman.undo_partition('partman_test.time_taptest_table', 20, p_target_table := 'partman_test.undo_taptest');
ERROR:  ALTER TABLE ... DETACH CONCURRENTLY cannot run inside a transaction block
CONTEXT: SQL statement " ALTER TABLE partman_test.time_taptest_table DETACH PARTITION partman_test.time_taptest_table_default CONCURRENTLY"
PL/pgSQL function partman.undo_partition(text,integer,text,boolean,numeric,text,text[],boolean) line 311 at EXECUTE
DETAIL: 
HINT: 
CONTEXT:  PL/pgSQL function partman.undo_partition(text,integer,text,boolean,numeric,text,text[],boolean) line 515 at RAISE
Time: 24.354 ms
```

keithf4:
Apologies I missed this reply earlier since it was on a closed issue. I think this may be a much bigger issue than just dropping a child table concurrently. I don't believe any operation that supports the concurrent flag can be done within a single transaction (CREATE INDEX, REINDEX, etc). So it's going to require some bigger changes in internals before something like this will be supported as far as I know.

jw1u1:
Don't run it within a Transaction Block. :-)
Running DETACH Non-Concurrently is not an option regarding our workloads.

keithf4:
All functions inherently run within an explicit transaction block in PostgreSQL. They all contain a `BEGIN` and `END` statement. Please see the linked issue. I tried doing this proposed change and it does not work.

keithf4:
> It could potentially be supported using dblink, even though setting up the credentials for it is rather ugly.

Yes, that is how the pg_jobmon extension that partman uses for logging works to monitor for errors - https://github.com/omniti-labs/pg_jobmon

PR #614, opened 2024-01-23
DETACH PARTITION CONCURRENTLY

jw1u1:
DETACH PARTITION CONCURRENTLY


keithf4:
Apologies but what is this PR for? It looks to be trying to merge something back from version 4.5.0?

If you're trying to add the feature to detach concurrently, that unfortunately cannot be done due to functions running within a transaction. See this issue https://github.com/pgpartman/pg_partman/issues/387

PR #618, opened 2024-01-26
Qualify partition_gap_fill with @extschema@

nicolagi:
I tried to install the extension to a new schema `partman5` so it can coexist with an older version under schema `partman` with commands like

    make
    sed -i -E 's/@extschema@/partman5/' sql/pg_partman--5.0.1.sql
    psql -U user db <sql/pg_partman--5.0.1.sql

but the install failed to create the function `partition_gap_fill` because it already exists.
I am _guessing_ that the function should be put in the `@extschema@` like other functions.

### DIFF
--- a/sql/functions/partition_gap_fill.sql
+++ b/sql/functions/partition_gap_fill.sql
@@ -1,4 +1,4 @@
-CREATE FUNCTION partition_gap_fill(p_parent_table text) RETURNS integer
+CREATE FUNCTION @extschema@.partition_gap_fill(p_parent_table text) RETURNS integer
     LANGUAGE plpgsql
     AS $$
 DECLARE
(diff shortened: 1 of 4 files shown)


keithf4:
Yes, that's correct. Good catch! I have a draft PR up for version 5.1 if you wouldn't mind doing your PR against that instead of the current master. It would need to be done on the two update files

https://github.com/pgpartman/pg_partman/pull/615/files#diff-a425e1b9be9404ca569c5fd5122710660cd7751b895ae5bf0b483bf85cb005a0R5

https://github.com/pgpartman/pg_partman/pull/615/files#diff-ba681fc479cb25728785b8b187be56bf22c4117349893873562262adb7623548R3384  (you may need to manually expand viewing this whole file to see it)

If you cannot, I'll have to apply the fix there so it gets put out in a proper update.

https://github.com/pgpartman/pg_partman/pull/615

nicolagi:
Sorry for the many force-pushes, that should be all the update files now! :pray:

keithf4 (approved):

PR #659, opened 2024-05-24
allow microsecond precision for epoch

agdolla:
This change adds microsecond precision to epoch, which allows seconds, milliseconds and nanoseconds but not microseconds right now.


keithf4:
Thank you for the PR! I'll get this added in during my next code review and it will be included in the next release.

Issue #660, opened 2024-05-27
Avoid `SELECT MAX(control)` queries


nicolagi:
I see that pg_partman runs `SELECT MAX(control)` to find what is the "next" partition to create.
That's expensive when a partition is large and there is no index on the control column (a situation I find myself in).
I wonder why not just trust the partition boundaries (perhaps only if `infinite_time_partitions` is true).
E.g., for a monthly-partitioned table `foo`, if `foo_p20240501` is present for values between first and last of May inclusive, assume new (monthly) partitions will be starting from `foo_p20240601` onwards.
I do see the need to still scan the DEFAULT table, in case there's recent data there that would conflict with new partitions.

IIRC, the current strategy also makes transitioning from monthly to, say, weekly partitions, hard. E.g., the `SELECT MAX(control)` method entails trying to create `foo_p20240508`, etc. which would conflict with the existing partition.

keithf4:
That's a good suggestion to revisit that method now that trigger-based partitioning is no longer supported. I'll definitely have a closer look during the next version's code review. Thank you!

Also curious on not having the index there. Is the control column purely for retention purposes and the queries never use it as a condition?

nicolagi:
> That's a good suggestion to revisit that method now that trigger-based partitioning is no longer supported. I'll definitely have a closer look during the next version's code review. Thank you!

Hopefully I'll be able to collaborate.

> Also curious on not having the index there. Is the control column purely for retention purposes and the queries never use it as a condition?

Precisely. (I was surprised as well at first.)

Forgive the possible off-topic, but it seems to me that if `run_maintenance` takes long, say, 5 minutes, then the initial DROP (when applying the retention policy) will acquire an exclusive lock on the table and block all queries to the table for the whole 5 minutes. If that reasoning is correct, it's all the more reason to avoid scanning if possible (and for me to index all control columns).

keithf4:
Just reviewing when the drop happens, I saw that it's one of the first steps in maintenance, so I'm going to review why that couldn't possibly be later instead so any locks the drop takes aren't done until closer to the end. I cannot recall if that lock is exclusive for the entire partition set or not. It's definitely exclusive for that child table tho.

This is also a use-case for using `run_maintenance_proc()` vs standard `run_maintenance()`. The procedure commits after each partition set's maintenance so one's transactional holds do not interfere with any others.

keithf4:
[...]

PR #662, opened 2024-05-30
Avoid scanning non-default partitions

nicolagi:
This is a minimal change to address #660.
It changes one query from selecting the max timestamp in a given partition (which may incur a full table scan) to selecting any timestamp (which should be fast).

**Testing.** I tested it with:

```
pg_prove -U $myuser -d $mydb -ovf test/*.sql
```

**Proof.** I'll argue that the query change does not change the behavior of the function.

Let P be the partition being iterated on.

If it's empty, then `v_max_timestamp` evaluates to null in both old and new code, and the loop iteration does nothing in both old and new code.

Let's then consider a non empty partition P.
Let A (for any timestamp) be the timestamp returned by the new code, and let M (for max) be that returned by the old code.
We have A <= M.

If A >= CURRENT_TIMESTAMP then `show_partition_name` gives back P for both old and new code. (Using A or M doesn't make a difference, as they're both in P.)

Let's look at the case when A < CURRENT_TIMESTAMP.
If also M < CURRENT_TIMESTAMP, both the old and the new code would use CURRENT_TIMESTAMP, so no difference.

It remans to consider the case

    A < CURRENT_TIMESTAMP <= M,

in which case CURRENT_TIMESTAMP still belongs to partition P, so the old code (uses M) and the new code (uses CURRENT_TIMESTAMP) both return P.

### DIFF
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -219,7 +219,7 @@ LOOP
         FOR v_row_max_time IN
             SELECT partition_schemaname, partition_tablename FROM @extschema@.show_partitions(v_row.parent_table, 'DESC', false)
         LOOP
-            EXECUTE format('SELECT max(%s)::text FROM %I.%I'
+            EXECUTE format('SELECT %s::text FROM %I.%I LIMIT 1'
                                 , v_partition_expression
                                 , v_row_max_time.partition_schemaname
                                 , v_row_max_time.partition_tablename
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -33,7 +33,7 @@ v_last_partition_timestamp      timestamptz;
 v_max_id                        bigint;
 v_max_id_default                bigint;
 v_max_time_default              timestamptz;
-v_max_timestamp                 timestamptz;
+v_child_timestamp               timestamptz;
 v_new_search_path               text;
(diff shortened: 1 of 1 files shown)


keithf4:
Thanks! Will take a look at this when I return to development.

sql/functions/run_maintenance.sql
```diff
@@ -219,24 +219,24 @@ LOOP
         FOR v_row_max_time IN
             SELECT partition_schemaname, partition_tablename FROM @extschema@.show_partitions(v_row.parent_table, 'DESC', false)
         LOOP
-            EXECUTE format('SELECT max(%s)::text FROM %I.%I'
+            EXECUTE format('SELECT %s::text FROM %I.%I LIMIT 1'
```
keithf4, line 222: Update the comments on lines 217 above and 241 below to be relevant that it's no longer getting the max value. Other than that, looks great! Thanks for the PR!

keithf4 (changes requested):

sql/functions/run_maintenance.sql
nicolagi: Quite right, I attempted to explain the new version, feel free to edit!

keithf4: Thanks!

keithf4 (approved):

PR #679, opened 2024-07-31
Fix link to CHANGELOG

mhagander:
It's now a MD file, not TXT.

(Yes, this is a drive-by fix, but I've clicked that link a few times now..)


keithf4:
Thank you!

PR #723, opened 2024-12-22
feat: process table columns with GENERATED ALWAYS for partition_data_id() and partition_data_time()

waitingsong:
columns with GENERATED ALWAYS and can not ignored by param `p_ignored_columns`, eg. PK
workaround:
insert `OVERRIDING SYSTEM VALUE` string into the sql if there is column having def `GENERATED ALWAYS` excepting `p_ignored_columns`.

### DIFF
--- a/sql/functions/table_has_identity_columns.sql
+++ b/sql/functions/table_has_identity_columns.sql
@@ -1,0 +1,40 @@
+CREATE OR REPLACE FUNCTION table_has_identity_columns(
+    table_name text,
+    p_ignored_columns text[] DEFAULT NULL
+) RETURNS boolean AS $$
+DECLARE
+    identity_columns_count int;
+    schema_name text;
+    table_only_name text;
+BEGIN
+    -- if table_name contains schema，split schema and table
+    IF table_name LIKE '%.%' THEN
+        schema_name := split_part(table_name, '.', 1);
+        table_only_name := split_part(table_name, '.', 2);
+    ELSE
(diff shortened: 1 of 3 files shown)


sql/functions/table_has_identity_columns.sql
```diff
@@ -0,0 +1,40 @@
+CREATE OR REPLACE FUNCTION table_has_identity_columns(
+    table_name text,
+    p_ignored_columns text[] DEFAULT NULL
+) RETURNS boolean AS $$
+DECLARE
+    identity_columns_count int;
```
keithf4: Appreciate this PR and will definitely look at getting it merged in when possible. Would you mind using the conventions I have in the other functions for parameter (prefix with `p_`)  and variable names (prefix with `v_`).

keithf4 (changes requested):

keithf4:
Just FYI, this fix will be in the next release, but I decided to do it a bit differently than you've done here to allow a little more flexibility. Thanks for your work, though, since it helped make things on my end easier to get done!

keithf4:
Apologies for the very long delay on this. I've included a modified version of this in the next release which has an RC1 tag out now. If you'd like to check it out and provide any feedback, I'd appreciate it.

https://github.com/pgpartman/pg_partman/releases/tag/v5.3.0RC1

keithf4:
5.3.0 has been released with this fix

PR #730, opened 2025-01-07
Implement encode/decode for ID columns

### LINKED ISSUE #729: Support for time-based integer IDs (e.g. snowflake) using encode/decode
Creating a new issue from https://github.com/pgpartman/pg_partman/issues/528#issuecomment-2572328140.

Currently the only supported ways to use IDs/integers in a control column are as a timestamp directly representing seconds, ms, us or ns since the UNIX epoch; or as a generic, non-timestamp ID range. #683 adds functionality to use ranges of UUID and text values via conversion functions. I want to add support for integers as well.

The primary example is "snowflake" IDs, used by [X](https://github.com/twitter-archive/snowflake/tree/b3f6a3c6ca8e1b6847baa6ff42bf72201e2c2231?tab=readme-ov-file#solution), [Discord](https://discord.com/developers/docs/reference#snowflakes), [Instagram](https://instagram-engineering.com/sharding-ids-at-instagram-1cf5a71e5a5c), [Mastodon](https://github.com/mastodon/mastodon/blob/5e796dc6f85b37c8378fe01cfd8ac23222c89eea/lib/mastodon/snowflake.rb#L30), and others. I brought them up in #388 and the resulting changes let me use snowflake IDs as ordinary integers, with offsets for interval and retention. This works well enough, but leads to gaps when no data is being added for a while since premake only looks at the current maximum value.

For background, snowflake values fit in 63 bits (for signed int64) and encode a timestamp plus a combination of worker and sequence information. Since each implementation uses a different bit range and epoch for the timestamp, partman won't be able to cover all possibilities with a single setting for `p_epoch` like 'snowflake'.

I propose a new option for p_epoch of 'func', which indicates that partman should use the encoder/decoder functions for IDs the same way it does for UUIDs and text. This should be a straightforward change, since most of the work has already been done in #683. Also, the parameter documentation for encoder/decoder mentions snowflakes, but those are only allowed if they are stored as a string rather than a bigint.

<details><summary>Example Discord ID encoder and decoder:</summary>

```sql
-- Discord snowflakes use a 41 bit timestamp and 22 bit worker/process/sequence.
-- The timestamp is the number of milliseconds since the first second of 2015 UTC (1420070400s since epoch)
CREATE FUNCTION discord_snowflake_to_timestamp(id bigint)
    RETURNS TIMESTAMPTZ LANGUAGE SQL IMMUTABLE PARALLEL SAFE AS
    $$SELECT TO_TIMESTAMP((id >> 22)/1000.0 + 1420070400)$$;

CREATE FUNCTION timestamp_to_discord_snowflake(ts TIMESTAMPTZ)
    RETURNS BIGINT LANGUAGE SQL IMMUTABLE PARALLEL SAFE AS
    $$SELECT CAST((EXTRACT(epoch FROM ts) - 1420070400)*1000 AS BIGINT) << 22$$;
```
</details>

calebj:
This PR allows users to specify the special value of 'func' for p_epoch to use custom functions to encode/decode time-ordered integers other than the classic seconds, ms, us or ns since the UNIX epoch.

Resolves #729.


sql/functions/run_maintenance.sql
```diff
@@ -18,7 +18,6 @@ v_analyze                       boolean := FALSE;
 v_check_subpart                 int;
 v_child_timestamp               timestamptz;
 v_control_type                  text;
-v_time_encoder                  text;
```
calebj, line 21: I removed this variable since it isn't used anywhere.

sql/functions/run_maintenance.sql
```diff
@@ -261,14 +262,23 @@ LOOP
...
 
+        -- avoids the need for a functional index using to_timestamp or decoder on control column to quickly find the max
+        v_max_control_expression := CASE
+            WHEN v_row.epoch = 'seconds' THEN format('to_timestamp(max(%I))', v_row.control)
+            WHEN v_row.epoch = 'milliseconds' THEN format('to_timestamp((max(%I)/1000)::float)', v_row.control)
+            WHEN v_row.epoch = 'microseconds' THEN format('to_timestamp((max(%I)/1000000)::float)', v_row.control)
+            WHEN v_row.epoch = 'nanoseconds' THEN format('to_timestamp((max(%I)/1000000000)::float)', v_row.control)
+            WHEN v_row.epoch = 'func' THEN format('%s(max(%I))', v_time_decoder, v_row.control)
+            ELSE format('max(%I)', v_row.control)
+        END;
```
calebj, line 273: This should remove the need for a functional index to find the maximum, provided there is an index on the underlying column. That should be a safe assumption to make since postgres prohibits creating a unique index or primary key on a partitioned table that doesn't contain the partition key.

If this works in all scenarios, it should also be ported to `partition_data_time()`

calebj:
Marking as draft because this still needs tests for smaller integer types

keithf4:
Thank you! Will be a bit busy over the next month or so, so may not get to dig into this for a bit.

Just reading your notes, though, it is possible to have unique indexes on non-control columns for individual tables using the template table system in partman. May also want to keep in mind that it may be possible in core in the future as well. So not sure if that will affect what you've done.

calebj:
Unique indexes on individual partitions or on non-control columns won't affect anything in this PR. The intent is to optimize the check by using `MAX()` on the base column before converting it to a timestamp, since otherwise it needs a functional index. I added that note because if there's no edge case where that would be undesirable, it makes sense to apply the same type of change to this section's calls to `MIN()` and `MAX()` https://github.com/pgpartman/pg_partman/blob/1192e933232a0ecc1b960470316f1fffa05a8da3/sql/functions/partition_data_time.sql#L131-L159

One other improvement I want to make is a round-trip and ordering sanity check for the passed functions, and an assert that the functions are at least marked `RETURNS NULL ON NULL INPUT` or `STRICT`, and `IMMUTABLE`.

Also, I think Julian dates might make an interesting example for 32-bit IDs, so I will play with that for a daily test.

rauanmayemir:
This PR made me realize I could adopt something like snowflake id and have time-based native partitioning for tables that need integer sequencial ids.

Looking forward to see this merged.

keithf4:
Been pretty busy lately and now getting back into pg_partman dev work. Just FYI I may not get this into the next release because it touches quite a bit and I have some other bug fixes I'd like to get in first and ensure they're all working fine independent of other big changes.. Will hopefully be able to look into this PR immediately after that.
Thanks again!

grodger:
Hi @calebj & @keithf4 
I was just wondering if and when this PR might be merged into the next release. I utilise snowflake quite extensively for large volumes of transactions so the 8 bytes vs 16 bytes for UUID 7 makes a material difference. To date I have built pg_partman with this PR incorporated, for my needs it is currently working well for the management of partitions but would obviously prefer it is mainlined.
Thanks,
Grant
PS. thanks for the awesome work that has gone into pg_partman

keithf4:
I didn't have this PR planned for the next release and I'm not sure I'll have the time to get it incorporated for it. I will definitely put it on the list of features to implement on the next release. Might even just do another update just for it since there's been more interest in it and due to the fact that I also work for Snowflake now myself. :)

grodger:
Thanks for the update @keithf4 !

rauanmayemir:
Gentle nudge to remind about this PR. 👀

keithf4:
Apologies. I will try and return to some of these open PRs in the near future. I had to get the other security fix out and trying to get a bug fix done now as well.

keithf4:
I'm working on the 5.5 release at the moment trying to get some other long standing issues wrapped up. This will be my next priority for 5.6

PR #752, opened 2025-02-19
Update pg_partman.md

fgrosse:
Just removing an outdated comment. It seems to me like the BGW now actually [does use `run_maintenance_proc`][1].

Thank you for the great extension!

[1]: https://github.com/pgpartman/pg_partman/blob/949bf56861d58ca8bf84d23ff5677025bfcf5bb6/src/pg_partman_bgw.c#L493

### DIFF
--- a/doc/pg_partman.md
+++ b/doc/pg_partman.md
@@ -417,7 +417,6 @@ run_maintenance_proc(
 ```
 
  * This procedure can be called instead of the `run_maintenance()` function to cause PostgreSQL to commit after each partition set's maintenance has finished. This greatly reduces contention issues with long running transactions when there are many partition sets to maintain.
- * NOTE: The BGW does not yet use this procedure and still uses the standard run_maintenance() function.
  * `p_wait` - How many seconds to wait between each partition set's maintenance run. Defaults to 0.
  * `p_analyze` - See p_analyze option in run_maintenance.
 


keithf4:
Thank you!

keithf4 (approved):

fgrosse:
I'm happy to contribute, even if its just small doc changes. I don't have permissions to merge the change, @keithf4,  will you take care of this?

I also have a question about `run_maintenance` but let me know if you would like me to copy it over to an issue instead: I noticed that `run_maintenance` has a `p_parent_table` argument but `run_maintenance_proc` doesn't. This currently prevents me from using the `_proc` version because I prefer to have control over which table I want to run the maintenance on. Do you think the argument could be added to the stored procedure as well?

keithf4:
> I'm happy to contribute, even if its just small doc changes. I don't have permissions to merge the change, @keithf4, will you take care of this?
> 

I'll take care of it.

> I also have a question about `run_maintenance` but let me know if you would like me to copy it over to an issue instead: I noticed that `run_maintenance` has a `p_parent_table` argument but `run_maintenance_proc` doesn't. This currently prevents me from using the `_proc` version because I prefer to have control over which table I want to run the maintenance on. Do you think the argument could be added to the stored procedure as well?

So the only thing the procedure does differently is run maintenance on more than one partition set and ensures there is a commit between each partition set to avoid a long running transaction.. It's just calling `run_maintnance()` in a loop. If you're only running it on one partition set, there is no need to use the procedure and you can just use the function directly.

fgrosse:
Thanks for the explanation :D

PR #755, opened 2025-02-27
fix update to 5.2.2

keithf4:
If user had started from 5.2.0, then the update to 5.2.2 and higher would not work


kmustkivi (approved):
The fix is confirmed to be working. Upgrade from 5.2.0 to 5.2.4 successful.

Issue #761, opened 2025-03-17
Setting `p_template_table` to false has unnecessary dependency on keeping `p_default_table` to `false`


mrusmanshahid:
I recently encountered an issue with Partman where I tried to set the `p_template_table=false` keeping `p_default_table` as default which means expecting the default table to be there but not the template itself. 

Instead of creating the partition, it gave me the following error
`Query 1 ERROR at Line 1: : ERROR:  No template table set in configuration for given parent table: public.test_table`

Upon investigation and reading the `create_parent` function I found out there is a block in there that expects it to be `false` 

```
ELSIF lower(p_template_table) IN ('false', 'f') THEN
    v_template_schema := NULL;
    v_template_tablename := NULL;
    RAISE DEBUG 'create_parent(): parent_table: %, skipped template table creation', p_parent_table;
```

Which is setting `v_template_schema` and `v_template_tablename` to `NULL` and following further in the code I found a block

```
IF p_default_table THEN
    -- Manage template inherited properties
    PERFORM partman.inherit_template_properties(p_parent_table, v_parent_schema, v_default_partition);
```

This expects the following function, `partman.inherit_template_properties` to run, and it's not behind any flag, as I highlighted above. Unfortunately, this function fails for obvious reasons, as the template table isn't there. 

In the end, setting `p_default_table to false`  (which skips the above block resulting in skipping the inherit template procedure call) fixed it but also eliminated the default partition. Ideally, these two configurations should be independent of each other.

keithf4:
Thanks for reporting that! Will try and get a fix for it in the next release

mrusmanshahid:
probably this should fix it, i checked the rest of the places and found out its behind the flag except this one place so this should potentially fix this problem --> https://github.com/pgpartman/pg_partman/pull/762

keithf4:
Thanks for reporting this and providing the fix! I'm working on the next update so I'll incorporate your PR into the work I already have started and make sure everything is covered.

keithf4:
I've tagged a release candidate for version 5.3.0 if you'd like to test out that this is working. Apologies for the delay!

https://github.com/pgpartman/pg_partman/releases/tag/v5.3.0RC1

keithf4:
5.3.0 has been released with this fix

PR #762, opened 2025-03-17
fix: put template inheritance behind conditional flag to avoid breaking the create parent function 

mrusmanshahid:
This pull-request should fix the issue highlighted here
https://github.com/pgpartman/pg_partman/issues/761


keithf4:
I've incorporated this change into an update file for version 5.3.0. Thank you!

Issue #767, opened 2025-04-11
To handle partition drop when Partition table has FK


AdarshHD:
https://github.com/pgpartman/pg_partman/blob/14114a36cf82a5ee7d177850e44d71e44d7a3761/sql/functions/drop_partition_time.sql#L259

In the above we are trying to drop the partition directly, if there is an foreign key referring partition table which we are trying to drop then we will start seeing failure which throws error saying that the partition is being referred.  Why can't we have approach like below 

- First detach the partition which is suppose to be dropped.  
- Secondly the drop the partition which it is suppose to drop.  

This way having foreign key referring the partitioned table doesn't make any issue when running maintenance.

somnam:
Bumped into the same issue. This seems to block from using pg_partman on partitioned tables with FK references when having `retention_keep_table = false` set.

Trying to drop a partitioned table with a FK reference fails with following error:
```
database-1  | 2025-09-01 09:13:58.494 UTC [839] ERROR:  cannot drop table parent_part_p20250501 because other objects depend on it
database-1  |    CONTEXT: SQL statement "DROP TABLE parent_part_p20250501"
database-1  |    PL/pgSQL function drop_partition_time(text,interval,boolean,boolean,text,timestamp with time zone) line 250 at EXECUTE
database-1  |    PL/pgSQL function partman.run_maintenance(text,boolean,boolean) line 336 at assignment
database-1  |    DETAIL: constraint fk_child_part__parent_id__parent_part on table child_part depends on table parent_part_p20250501
database-1  |    HINT: Use DROP ... CASCADE to drop the dependent objects too.
database-1  |    CONTEXT: PL/pgSQL function drop_partition_time(text,interval,boolean,boolean,text,timestamp with time zone) line 307 at RAISE
database-1  |    PL/pgSQL function partman.run_maintenance(text,boolean,boolean) line 336 at assignment
database-1  |    DETAIL:
database-1  |    HINT:
database-1  | 2025-09-01 09:13:58.494 UTC [839] CONTEXT:  PL/pgSQL function partman.run_maintenance(text,boolean,boolean) line 474 at RAISE
database-1  | 2025-09-01 09:13:58.494 UTC [839] STATEMENT:  SELECT partman.run_maintenance(p_parent_table => 'parent_part', p_analyze => true);

```
The table has:
```
Referenced by:
[...]

PR #768, opened 2025-04-11
Commiting Changes to handle partition maintenance on Partition table …

AdarshHD:
Earlier in 4.7.1 we had column in part_config table drop_cascade_Fk, so if the partitioned table on which maintenance was run using run_maintenance, if it had any Reference then those things use to get dropped. But that was not right, I guess that reason it has been removed from 5.1.0 version.  But a new problem comes if partitioned table is being referred then run maintenance will fail. In order to handle this we need to detach partition before dropping it, hence I have made the changes in two place where partition was being dropped, this should work smoothly without any issue even when partition table is being referred by another table.   

had reported this as issue in https://github.com/pgpartman/pg_partman/issues/767

### DIFF
--- a/sql/functions/drop_partition_id.sql
+++ b/sql/functions/drop_partition_id.sql
@@ -248,6 +248,12 @@ LOOP
                 IF v_jobmon_schema IS NOT NULL THEN
                     v_step_id := add_step(v_job_id, format('Drop table %s.%s', v_row.partition_schemaname, v_row.partition_tablename));
                 END IF;
+				v_sql := format('ALTER TABLE %I.%I DETACH PARTITION %I.%I'
+                , v_parent_schema
+                , v_parent_tablename
+                , v_row.partition_schemaname
+                , v_row.partition_tablename);
+				EXECUTE v_sql;
                 v_sql := 'DROP TABLE %I.%I';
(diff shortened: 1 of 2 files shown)


keithf4:
This looks good but would like to get a unit test in place to ensure it works as expected now and in the future. If you have a chance to do one in the near future, then I can immediately merge this in as a bug fix and get a patch release out. 

Otherwise I'll look at getting that unit test done and in with 5.4 when that release is ready.

keithf4 (changes requested):
EDIT: meant to do this comment on the Issue

keithf4:
Pushing this to 5.5 for now since I didn't get a chance to write a unit test for it

AdarshHD:
Note :  I just wanted to bring to attention that when we detach partition if the table has FK it will run data validation this will take time if the table is having huge  data, I have bumped into this in my production.  So we need to do better testing around this.  
We can use concurrently but if we have default partition even that is not allowed in PgSQL, So need brainstorming  around this.

keithf4:
Merging this in and going to see about making it optional via a config flag. It will default to disabled so the users will get an error vs running into the performance issues. The documentation for the flag will note the potential performance issues around doing the detached operation as part of normal maintenance

PR #780, opened 2025-05-21
Commit last transaction before ANALYZE phase

marcus-kempe:
The way the code works prior to this change proposal, is that the transaction open for the final table to have constraints applied, will be left un-commited during the ANALYZE phase.  Since an ACCESS EXCLUSIVE lock will still be held, this can be problematic.


keithf4:
Good catch. Thank you!

keithf4 (approved):

PR #798, opened 2025-08-19
prevent deadlocks by run_maintenance

jw1u1:
run_maintenance locks partitions first. This is opposite to common table access and causes deadlocks.
This patch locks the parent table before the partitions are accessed.

### DIFF
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -227,7 +227,7 @@ LOOP
         -- Avoids doing a scan on entire partition set and/or getting any values accidentally in default.
         -- Lock the parent_table first to prevent deadlocks
         RAISE DEBUG 'PERFORM LOCK %.%', v_parent_schema, v_parent_tablename;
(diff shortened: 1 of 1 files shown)


jw1u1:
We still face deadlocks.
I'm going to create a new PR, if this is fixed.

keithf4:
Just curious if you were able to find a better locking pattern in maintenance? Looks like someone else may be having the same issue.

https://github.com/pgpartman/pg_partman/discussions/833

jw1u1:
Hi @keithf4 ,
regardless in which order i manually locked parent/paritions, i didn't get rid of the deadlocks.
I'm not able to reproduce it, it's just one one productive DB.

keithf4:
Ok. Since you narrowed it down to being a conflict between the locks on the default and the parent, this seems more like a possible bug or edge case in core and not something I'd really be able to control in pg_partman. If you do find any work-arounds or fixes for this, please let me know. The only solution I'd come up with so far was maybe dropping the default table if it wasn't needed.

PR #799, opened 2025-08-19
Warn about Extension Version mismatch

### LINKED ISSUE #735: Feature Request: check version mismatch
Hi Keith,

some partman functions like run_maintenance silently do nothing, if the loaded shared_preload_library is a newer version than the currently installed version in the database.
I would be helpful to have clear error messages in this case.

Rgds, Jens

jw1u1:
fixes 50%  #735

### DIFF
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -98,7 +98,7 @@ EXECUTE format('SELECT set_config(%L, %L, %L)', 'search_path', v_new_search_path
 
 SELECT true INTO v_version_mismatch FROM pg_available_extensions WHERE name = 'pg_partman' AND default_version != installed_version;
 IF v_version_mismatch THEN
-    RAISE WARNING 'A new version of pg_partman is available. run_maintenance might nor work as expected. Run ALTER EXTENSION pg_partman UPDATE first.';
+    RAISE WARNING 'A new version of pg_partman is available. run_maintenance might not work as expected. Run ALTER EXTENSION pg_partman UPDATE first.';
 END IF;
 
 IF v_jobmon_schema IS NOT NULL THEN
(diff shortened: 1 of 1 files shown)


sql/functions/run_maintenance.sql
```diff
@@ -225,6 +231,9 @@ LOOP
 
         -- Loop through child tables starting from highest to get a timestamp from the highest non-empty partition in the set
         -- Avoids doing a scan on entire partition set and/or getting any values accidentally in default.
+        -- Lock the parent_table first to prevent deadlocks
+        RAISE DEBUG 'PERFORM LOCK %.%', v_parent_schema, v_parent_tablename;
```
keithf4: Please keep this PR separate from the version check PR

sql/functions/run_maintenance.sql
```diff
@@ -95,6 +96,11 @@ IF p_jobmon THEN
 END IF;
 EXECUTE format('SELECT set_config(%L, %L, %L)', 'search_path', v_new_search_path, 'false');
 
+SELECT true INTO v_version_mismatch FROM pg_available_extensions WHERE name = 'pg_partman' AND default_version != installed_version;
+IF v_version_mismatch THEN
+    RAISE WARNING 'A new version of pg_partman is available. run_maintenance might not work as expected. Run ALTER EXTENSION pg_partman UPDATE first. A restart might be required';
+END IF;
```
keithf4, line 103: Probably want to adjust either the check or the WARNING value a little bit because this would also throw an error if the default version somehow got set to a lower value than what is installed. So can either do a > check for the `true` value or adjust the WARNING to note that there is simply a mismatch between the default version and what is installed. 

I think I'd prefer the latter and also include the default and installed version values in the WARNING message labelled appropriately.

keithf4 (changes requested):

sql/functions/run_maintenance.sql
jw1u1: done

sql/functions/run_maintenance.sql
jw1u1: Sorry, removed.

keithf4 (approved):

keithf4:
I've pushed the update for v5.3.1 up to the development branch that includes this fix as well as a check for the matching shared library file. I'll be trying to tag and release this next week if you'd like to test it in the mean time

PR #805, opened 2025-09-09
improved exception handling, skip tables that are causing errors

jw1u1:
Incorrectly configured tables have previously led to the termination of the procedure
Untested for id based partitioning!

### DIFF
--- a/sql/functions/run_maintenance.sql
+++ b/sql/functions/run_maintenance.sql
@@ -18,6 +18,7 @@ v_analyze                       boolean := FALSE;
 v_check_subpart                 int;
 v_child_timestamp               timestamptz;
 v_control_type                  text;
+v_exception                     boolean;
 v_time_encoder                  text;
 v_time_decoder                  text;
 v_create_count                  int := 0;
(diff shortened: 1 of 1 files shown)


keithf4 (approved):

keithf4:
So this was a noble attempt to try and catch exceptions and move on to other partition sets, but there's far too many places that are going to be needed to try and catch issues. Testing this myself I kept running into more and more places that would need to be caught, especially when calling other functions. Rather than have some of these caught and not others, I'd rather not have this being handled at all so there's not some expectation that all the other places need to be caught and handled as well.

Thank you for the attempt, but I'll be undoing this commit on my end.

keithf4:
If there is some issue with partition maintenance not working with a specific table, the `automatic_maintenance` column in `part_config` can be set to `off` and the calls to `run_maintenance()` with no argument will skip those partition sets.

jw1u1:
Rejecting this patch is not a valid attempt to make partman smarter @keithf4 
It's not only config Erros but also Errors like this:
fcdb=> select partman.run_maintenance('fc.rawdata');
ERROR:  cache lookup failed for constraint 603899273

PR #817, opened 2025-09-27
Refine tag matching in the release workflow

theory:
Improve the `on.push.tags` entry in `.github/workflows/release.yml` to use a regular expression that matches only non-pre-release semantic version tags. This should prevent it from releasing a prerelease on PGXN. It will need refining should the desire arise to put prerelease versions on PGXN.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,7 @@
 name: Release
 on:
   push:
-    tags: [v*]
+    tags: ["v[0-9]+.[0-9]+.[0-9]+"]
 jobs:
   release:
     name: Release on PGXN


theory:
Actually, based on my reading of [filter patterns](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet), it should probably be:

```yaml
    tags:
      - "v[0-9]+.[0-9]+.[0-9]+"
      - "![-]"
```

The _believe_ that the second pattern will exclude anything matched by the first that also includes a dash, presuming you're using [semantic versions](https://semver.org) where prerelease version info appears after a dash.

keithf4 (approved):

PR #829, opened 2025-11-05
Fix assignment operator bug in DESC order partition data functions

### LINKED ISSUE #828: p_order := DESC causes partition_data_proc to fail
I set up a parent table with monthly partitions:
```
SELECT partman.create_parent(
 p_parent_table => 'my_table',
 p_control      => 'start_timestamp',
 p_type         => 'range',
 p_interval     => '1 month',
 p_premake      => 3);
```

That creates partitions as expected (today is 2025-11-05):
```
Partitions: my_table_p20250801 FOR VALUES FROM ('2025-08-01 00:00:00+00') TO ('2025-09-01 00:00:00+00'),
            my_table_p20250901 FOR VALUES FROM ('2025-09-01 00:00:00+00') TO ('2025-10-01 00:00:00+00'),
            my_table_p20251001 FOR VALUES FROM ('2025-10-01 00:00:00+00') TO ('2025-11-01 00:00:00+00'),
            my_table_p20251101 FOR VALUES FROM ('2025-11-01 00:00:00+00') TO ('2025-12-01 00:00:00+00'),
            my_table_p20251201 FOR VALUES FROM ('2025-12-01 00:00:00+00') TO ('2026-01-01 00:00:00+00'),
            my_table_p20260101 FOR VALUES FROM ('2026-01-01 00:00:00+00') TO ('2026-02-01 00:00:00+00'),
            my_table_p20260201 FOR VALUES FROM ('2026-02-01 00:00:00+00') TO ('2026-03-01 00:00:00+00'),
            my_table_default DEFAULT
```

Then I can successfully call `partition_data_proc` with a `p_interval` less than the partition size:
```
CALL partman.partition_data_proc(
 p_parent_table := 'my_table',
 p_source_table := 'my_table_nonpartitioned',
 p_interval     := '1 hour');
```

**But if I instead try the same call using `p_order := DESC`, this call fails:**
```
CALL partman.partition_data_proc(
 p_parent_table := 'my_table',
 p_source_table := 'my_table_nonpartitioned',
 p_interval     := '1 hour',
 p_order        := 'DESC');
```

**With this error:**
```
ERROR:  relation "my_table_p20251105" does not exist
```

Note that it's attempting to insert data into a partition (`20251105`) that doesn't exist. It seems to be basing that on the current date / the "beginning" of the DESCENDING ordering -- instead of basing it on the MONTH START of that date.

mattklein:
Fixes #828

  ## Problem
  The `partition_data_proc` function failed when called with `p_order := 'DESC'` and a `p_interval` parameter, causing errors like:
  ERROR: relation "my_table_p20251105" does not exist

  ## Root Cause
  The code incorrectly used `=` (comparison operator) instead of `:=` (assignment operator) in PL/pgSQL. This caused variable assignments to be evaluated as boolean comparisons and discarded, rather than actually updating the
  variables.

  ## Solution
  Changed all affected instances of `=` to `:=` for proper variable assignments in:
  - `partition_data_time.sql` line 252: `v_min_partition_timestamp` assignment for DESC order with batch intervals
  - `partition_data_id.sql` lines 154, 168, 174: `v_min_partition_id` assignments for both ASC and DESC order

  ## Testing
  The fix ensures that when using DESC order, the partition boundary calculations correctly respect the partition interval (e.g., monthly) rather than defaulting to incorrect daily boundaries.

### DIFF
--- a/sql/functions/partition_data_id.sql
+++ b/sql/functions/partition_data_id.sql
@@ -151,7 +151,7 @@ FOR i IN 1..p_batch_count LOOP
         IF v_start_control IS NULL THEN
             EXIT;
         END IF;
-        v_min_partition_id = v_start_control - (v_start_control % v_partition_interval);
+        v_min_partition_id := v_start_control - (v_start_control % v_partition_interval);
         v_partition_id := ARRAY[v_min_partition_id];
         -- Check if custom batch interval overflows current partition maximum
         IF (v_start_control + p_batch_interval) >= (v_min_partition_id + v_partition_interval) THEN
@@ -165,13 +165,13 @@ FOR i IN 1..p_batch_count LOOP
         IF v_start_control IS NULL THEN
             EXIT;
         END IF;
-        v_min_partition_id = v_start_control - (v_start_control % v_partition_interval);
+        v_min_partition_id := v_start_control - (v_start_control % v_partition_interval);
         -- Must be greater than max value still in parent table since query below grabs < max
         v_max_partition_id := v_min_partition_id + v_partition_interval;
         v_partition_id := ARRAY[v_min_partition_id];
         -- Make sure minimum doesn't underflow current partition minimum
         IF (v_start_control - p_batch_interval) >= v_min_partition_id THEN
-            v_min_partition_id = v_start_control - p_batch_interval;
+            v_min_partition_id := v_start_control - p_batch_interval;
         END IF;
     ELSE
         RAISE EXCEPTION 'Invalid value for p_order. Must be ASC or DESC';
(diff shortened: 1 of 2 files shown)


mattklein:
NOTE: PR generated by Claude Code based on the issue I'd written up [here](https://github.com/pgpartman/pg_partman/issues/828). Admittedly I'm not a PL/SQL expert so I'm not sure this is a valid fix, and I haven't build/tested this myself, but submitting it in the hopes it'll be useful to the maintainers. 

Btw I've loved my first experience with pg_partman!

keithf4 (approved):

keithf4:
Thank you!

PR #830, opened 2025-11-10
Refactor partition suffix logic to allow custom child prefixes

pavan-postgres:
Updated the logic for determining suffixes of partitioned tables. 

Previous implementation always appended '_p' for partitioned tables:

    v_suffix := format('%s%s', CASE WHEN p_table_partition THEN '_p' END, p_suffix);

New logic allows using a simple naming convention for child tables without affecting existing behavior:

    v_suffix := format('%s%s',
        CASE
            WHEN p_table_partition AND p_simple_naming THEN '_'
            WHEN p_table_partition THEN '_p'
        END,
        p_suffix
    );

This change enables specifying a custom prefix/suffix for child table names (via `p_suffix`) while preserving the original '_p' behavior for partitions that do not use simple naming. It makes the code more flexible without breaking existing functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+5.6.0
+=====
+NEW FEATURES
+------------
+ - Allow a custom prefix to be used between the parent table name and the partition suffix when naming child tables, instead of the hardcoded `_p`. A new `part_config` column, `child_table_prefix`, controls this and defaults to `_p` to preserve existing behavior for all current partition sets. Set it with `UPDATE part_config SET child_table_prefix = '...' WHERE parent_table = ...` after the partition set is created (like other options not exposed directly on `create_parent()`/`create_partition()`). Takes effect on all subsequently created child tables (`run_maintenance()`, `create_partition_time()`/`create_partition_id()`, `partition_data_time()`/`partition_data_id()`, and `show_partition_name()`); existing child tables are not renamed.
+
+
 5.5.0
 =====
 BREAKING CHANGES
--- a/META.json
+++ b/META.json
@@ -1,7 +1,7 @@
 {
     "name": "pg_partman",
     "abstract": "Extension to manage partitioned tables by time or ID",
-    "version": "5.5.0",
+    "version": "5.6.0",
     "maintainer": [
         "Keith Fiske <keith@keithf4.com>"
     ],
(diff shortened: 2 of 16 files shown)


keithf4:
Don't think I'm going to be able to get this into the next release, but I do like the idea. Will come back to review this after next release is out. Thank you!

keithf4:
Apologies for the delay in response. Reviewing this again, so this doesn't really allow custom prefixes to the child name suffix, it just allows the removal of the "p" in the suffix name. It also doesn't boil up to a point where it actually allows the user to put the setting into effect. So on second thought I'm not sure I'd like to implement this as is for now.

If maybe you wanted to redo this to actually allow full customization of the prefix to the childname suffix naming pattern, I may consider that. Although I am a little hesitant on adding more flags to the create_partition() function right now.

Can you perhaps more clearly state the intended usage of this?

pavan-postgres:
Sorry for the slow follow-up, and thanks for the detailed feedback — both points are fair, so I reworked this instead of just patching around them.

**"Doesn't really allow custom prefixes"** — agreed, `p_simple_naming` was just a boolean toggle between `_p` and `_`. I've replaced it with `check_name_length(..., p_partition_prefix text DEFAULT '_p')`, so the prefix is now an arbitrary string, not a fixed choice of two.

**"Doesn't boil up to where the user can actually use it"** — also agreed. It's now backed by a real `part_config.child_table_prefix` column (defaults to `_p`, so existing partition sets are untouched) and threaded through every place a child table name gets generated: `create_partition_time()`, `create_partition_id()`, `show_partition_name()`, `partition_data_time()`, `partition_data_id()`, and `dump_partitioned_table_definition()`.

On the "hesitant to add more flags to `create_partition()`" point — I deliberately did *not* add a parameter there. Like `retention`, `optimize_constraint`, etc., it's set after the fact:

```sql
UPDATE part_config SET child_table_prefix = '_' WHERE parent_table = 'my_schema.my_table';
```

Intended usage: anyone who wants child tables named `my_table_20260101` instead of `my_table_p20260101`, or a fully custom convention like `my_table_batch_200`, sets this once and it applies to every child table created after that (initial premake, `run_maintenance()`, manual `create_partition_time()`/`create_partition_id()` calls, and data-movement functions).

Scope-wise I kept this to top-level partition sets only (not `part_config_sub`/subpartitioning) to keep the diff small and reviewable, given subpartitioning is already fairly delicate code — happy to add sub-partition support as a follow-up if that's useful.

Added pgTAP coverage(took Claude's help) for both time- and id-based partitioning, a migration script, and doc/changelog updates. Ran the full test suite locally (PG17, `pgxn/pgxn-tools`, same as CI) — 4198/4198 passing.

Issue #836, opened 2026-01-14
Security vulnerability with call to array_to_string


ujjwalpl:
pg_partman SQL extension code calls array_to_string without specifying the namespace pg_catalog. An attacker can escalate privileges by defining own array_to_string SQL function that has arbitrary code in it when postgresql server calls the defined malicious function. We need to fix the pg_partman extension SQL definitions so that calls to array_to_strings are changed to pg_catalog.array_to_strings.

keithf4:
Good catch. I tried to be thorough with built-in catalog calls but seems I missed some. If you find any others, please let me know!

keithf4:
Seems I actually missed this on a lot of functions. Was pretty good when querying catalog views and tables, but missed a lot of common functions. Will work on the next update to be a patch to fix as many of these as I can find. Thank you!

A note of some relief, though, is that none of the functions are security definer. So the functions can only do things that the user themselves has privileges to do. So as it is now, there should be no worry of privilege escalation that I can see. 

One big thing that you can do with pg_partman, and is the default recommended, is to install it to its own schema. Then only give specific users privileges to use the schema that partman is in. That can help further mitigate issues.

keithf4:
PR #837 should hopefully address this and all search path related issues going forward. Thank you again for bringing this up. Made me re-evaluate how I handled search paths throughout the extension and PG functions coding going forward!

keithf4:
If you have time to test this PR, I would appreciate it. I'm hoping to get this released in the next week or so.

keithf4:
This has been fixed with version 5.4.1

PR #837, opened 2026-01-22
v5.4.1 ensure secure search paths in all functions and procedures

keithf4:
Fix security vulnerability brought up in #836. Fixes the ones mentioned there with hopefully all other concerns around search paths in all functions and procedures.

Fix handling timestamp without timzone

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+5.4.1
+=====
+BUGFIXES
+--------
+ - Ensure all functions have a search path set to avoid security issues around user object overrides. Ensure all procedures have all object calls schema qualified since a default search path cannot be set in them. (Github Issue #836)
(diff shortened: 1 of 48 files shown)


spencerbryson:
@keithf4 just an FYI, I didn't want to create a new issue for this, but I noticed format usage is still unqualified within most functions.

BradleyLi:
@keithf4 .format() has not been qualified so people can still write their own .format() and cause naming conflicts, i.e.  they happen to have their own format function in the first schema and cause it to throw error.

keithf4:
> @keithf4 .format() has not been qualified so people can still write their own .format() and cause naming conflicts, i.e. they happen to have their own format function in the first schema and cause it to throw error.

All functions (not procedures) should have a SET search path declaration that is active for the full run of the function. If you see any missing please let me know

keithf4:
Oh i see what you mean that if they put it in the extension's schema. That should still be able to be handled by locking down the permissions on the schema so not just anyone can make an object in that schema

Issue #848, opened 2026-02-24
5.4.2 pg_partman.control file not updated to 5.4.2


jreidthompson:
The 5.4.2 pg_partman.control file lists default version 5.4.1

 ~/bin/pgsql/pg_partman-5.4.2
$ cat pg_partman.control
default_version = '5.4.1'
comment = 'Extension to manage partitioned tables by time or ID'
relocatable = false
superuser = false
requires = 'plpgsql'

jreidthompson:
Sorry - meant to note this is  in the zip file

jreidthompson:
just verified still 5.4.1 in development branch

keithf4:
Thank you. I'll actually try and get a new version pushed out since the packages that were built now have the wrong control file in them.

keithf4:
Version 5.4.3 has been released. Should resolve any issues about this missing control file version.

PR #850, opened 2026-03-02
fix: allow toast relopt inheritance from template

### LINKED ISSUE #812: Inherit template toast table's properties while creating partition.
As listed in https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-STORAGE-PARAMETERS, In postgres we can set toast related storage parameter while creating or altering table. But when doing the same for template table the toast specific properties are not inherited when new paritition is being created.

This is happening because we are only copying reloptions from template table pg_class entry, whereas toast specific storage parameter are stored in reloptions for the pg_toast table of the respective template table.

keithf4:
Fixes #812 & #848

### DIFF
--- a/sql/functions/inherit_template_properties.sql
+++ b/sql/functions/inherit_template_properties.sql
@@ -25,6 +25,7 @@ v_template_schemaname   text;
 v_template_table        text;
 v_template_tablename    name;
 v_template_unlogged     char;
+v_toast_table_oid       oid;
 
 BEGIN
 /*
@@ -212,6 +213,26 @@ LOOP
     RAISE DEBUG 'inherit_template_properties: Set relopts: %', v_sql;
     EXECUTE v_sql;
 END LOOP;
+
+-- Get toast table options
+SELECT reltoastrelid INTO v_toast_table_oid
+FROM pg_catalog.pg_class c
+JOIN pg_catalog.pg_namespace n ON c.relnamespace = n.oid
+WHERE c.oid = v_template_oid;
+
+FOR v_relopt IN
+    SELECT unnest(reloptions) as value
+    FROM pg_catalog.pg_class
+    WHERE oid = v_template_oid
+LOOP
+    v_sql := format('ALTER TABLE %I.%I SET (toast.%s)'
+                    , v_child_schema
+                    , v_child_tablename
+                    , v_relopt.value);
+    RAISE DEBUG 'inherit_template_properties: Set toast relopts: %', v_sql;
+    EXECUTE v_sql;
+END LOOP;
+
 RETURN true;
 
 END
--- a/sql/functions/inherit_template_properties.sql
+++ b/sql/functions/inherit_template_properties.sql
@@ -223,7 +223,7 @@ WHERE c.oid = v_template_oid;
 FOR v_relopt IN
     SELECT unnest(reloptions) as value
     FROM pg_catalog.pg_class
-    WHERE oid = v_template_oid
+    WHERE oid = v_toast_table_oid
 LOOP
     v_sql := format('ALTER TABLE %I.%I SET (toast.%s)'
                     , v_child_schema
--- a/test/test-id-gap-fill.sql
+++ b/test/test-id-gap-fill.sql
@@ -27,7 +27,7 @@ CREATE INDEX ON partman_test.id_taptest_table (col3);
 
 -- TODO Add checks that these settings are in place on child tables
 ALTER TABLE partman_test.template_id_taptest_table SET (autovacuum_vacuum_threshold = 100);
-ALTER TABLE partman_test.template_id_taptest_table SET (toast.autovacuum_vacuum_threshold = 100);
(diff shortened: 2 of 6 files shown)


sql/functions/inherit_template_properties.sql
```diff
@@ -212,6 +213,26 @@ LOOP
...
+-- Get toast table options
+SELECT reltoastrelid INTO v_toast_table_oid
+FROM pg_catalog.pg_class c
+JOIN pg_catalog.pg_namespace n ON c.relnamespace = n.oid
+WHERE c.oid = v_template_oid;
+
+FOR v_relopt IN
+    SELECT unnest(reloptions) as value
+    FROM pg_catalog.pg_class
+    WHERE oid = v_template_oid
```
chopraanmol1: Replace `v_template_oid` with `v_toast_table_oid`

keithf4: Thanks! Glad I asked in addition to working on the unit test for it today.

keithf4: This has been fixed

test/test-id-gap-fill.sql
```diff
@@ -24,6 +25,10 @@ CREATE TABLE partman_test.template_id_taptest_table (LIKE partman_test.id_taptes
 ALTER TABLE partman_test.id_taptest_table ADD PRIMARY KEY (col1);
 CREATE INDEX ON partman_test.id_taptest_table (col3);
 
+-- TODO Add checks that these settings are in place on child tables
+ALTER TABLE partman_test.template_id_taptest_table SET (autovacuum_vacuum_threshold = 100);
```
chopraanmol1: Consider using a different relopt for the parent table from toast table. Right now a potential bug where partman is applying both parent tables & it's toast tables rel option to child's toast table will not fail this test. 

for parent table reloption using something like `fillfactor` might even be better because `toast.fillfactor` is not a valid reloption so in case the above mentioned potential bug gets introduced we will see failure from inherit_template_properties

test/test-id-gap-fill.sql
```diff
@@ -54,6 +59,61 @@ SELECT is_empty('SELECT * FROM ONLY partman_test.id_taptest_table', 'Check that
 SELECT results_eq('SELECT count(*)::int FROM partman_test.id_taptest_table', ARRAY[9], 'Check count from parent table');
 SELECT results_eq('SELECT count(*)::int FROM partman_test.id_taptest_table_p3000000000', ARRAY[9], 'Check count from id_taptest_table_p3000000000');
 
+-- Check reloptions set for child table
+SELECT results_eq('WITH relopt AS (
+    SELECT unnest(reloptions) AS relopt FROM pg_class c
```
chopraanmol1, line 63: Do we need `unnest` here? 

Currently we are testing only if the correct relopts is applied, we should ideally also test if no additional relopts is applied. Removing unnest would mean we can compare all of the relopts instead of just one.

chopraanmol1 (commented):
Functionality point of view the new changes are working correctly for my use case.

test/test-id-gap-fill.sql
keithf4: Good suggestion, thank you.

test/test-id-gap-fill.sql
keithf4: I'm not sure if this is really necessary. If someone sets other relopts other than the ones set via the template table, that's kind of on them and I don't see any reason to disallow it nor be that concerned.

If you'd like to adjust the test to be more flexible to test for additional options like that, I wouldn't mind and would be happy to accept the PR.

keithf4: Also just want to say thank you for your time and detailed PR reviews!

chopraanmol1: I was not referring to user setting up additional relopts. My comment was more to ensure that we have test which make sures partman does not inherit relopts from other pg class entries. Right now with fillfactor relopts we have ensured parent table's relopts are not copied to child toast table's relopts, but nothing is verifying if parent toast relopts(or some other table's relopts) is not being copied to child table relopts.

If you feel it's not really needed right now then feel free to resolve this thread, rest of the MR already looks good. Also thanks for picking up the changes needed to resolve this issue.

keithf4: I see. But since we're copying the relopts from a specific OID entry in pg_class, I'm curious how relopts from another pg_class entry could be obtained inadvertently? Genuinely curious if this is something you've seen before and should be a concern when doing catalog lookups like this.

chopraanmol1: Currently I don't think that will happen as the query looks correct. The suggestion was more towards ensuring that any future changes in partman does not introduce regression where slightly incorrect pg_class query or ALTER query starts copying additional relopts. In an ideal case the tests should be able to catch regression like this, but as I mentioned if you feel it's not needed immediately then we can resolve the thread.

chopraanmol1 (approved):

Issue #865, opened 2026-05-08
inconsistent timezone handling in case of retention


kurjam:
Timezone calculation of partition boundaries is off when dealing with non-gmt timezones when retention is set.

pg_partman version 5.3.1
postgresql 18.1

Although it might seem like an edge case, our different services have stumbled on this multiple times. The problem is that all sorts of client drivers set the timezone of a session to GMT silently while the db server timezone is something else. So when the pg_partman_bgw maintenance runs it does so in server time. 

This is an example from one of our dbs:

[local]:5432 postgres@test> \d+ public.test

Partitions: 
            test_p20260501 FOR VALUES FROM ('2026-04-30 19:00:00-05') TO ('2026-05-31 19:00:00-05'),
            test_p20260601 FOR VALUES FROM ('2026-05-31 19:00:00-05') TO ('2026-06-30 19:00:00-05'),
            test_p20260701 FOR VALUES FROM ('2026-06-30 19:00:00-05') TO ('2026-07-31 19:00:00-05'),
            test_p20260801 FOR VALUES FROM ('2026-07-31 19:00:00-05') TO ('2026-08-31 19:00:00-05'),
            test_p20260901 FOR VALUES FROM ('2026-08-31 19:00:00-05') TO ('2026-09-30 19:00:00-05')

These boundaries were being treated just fine (e.g. new partitions created by run_maintenance()) when we initially deployed this service on pg_partman 4.7.0. Now after upgrading to 5.3.1 things have gone downhill.

If the retention setting is on, subsequent partitions created by pg_partman will look like this:

[local]:5432 postgres@test> \d+ public.test
Partitions: 
            test_p20260501 FOR VALUES FROM ('2026-04-30 19:00:00-05') TO ('2026-05-31 19:00:00-05'),
            test_p20260601 FOR VALUES FROM ('2026-05-31 19:00:00-05') TO ('2026-06-30 19:00:00-05'),
            test_p20260701 FOR VALUES FROM ('2026-06-30 19:00:00-05') TO ('2026-07-31 19:00:00-05'),
            test_p20260801 FOR VALUES FROM ('2026-07-31 19:00:00-05') TO ('2026-08-31 19:00:00-05'),
            test_p20260901 FOR VALUES FROM ('2026-08-31 19:00:00-05') TO ('2026-09-30 19:00:00-05'),
[...]

PR #872, opened 2026-07-17
fix: align time partition boundaries to the set's timezone (#865)

x4m:
run_maintenance() did timestamptz month arithmetic in the session timezone, so a partition set aligned to a non-UTC zone drifted when maintenance ran from a different session (e.g. a UTC client such as DataGrip): new boundaries and child names no longer landed on the intended day.

The timezone a set is aligned to is now recorded at creation in the new part_config.partition_timezone column (overridable via p_timezone on create_parent/create_partition), and all boundary math is done in that timezone. A NULL value keeps the previous session-timezone behavior, so existing sets are unaffected until it is set. Adds an upgrade script and a pgTAP regression test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+5.4.4
+=====
+BUGFIXES
+--------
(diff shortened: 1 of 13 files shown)


x4m:
Still draft, not ready for review yet. Sorry for posting too  early.

keithf4:
I'm not sure I'm in favor of this feature quite yet. I've already documented how timezones are expected to work based on the client  that runs the commands

 https://github.com/pgpartman/pg_partman/blob/development/doc/pg_partman.md#time-zones

x4m:
Thanks for the pointer! I need to think about it. Should I close the PR so far, or let's keep it for a while?

keithf4:
You can leave it open for now if you'd like in case there's an idea you come up with around it that you'd like to share.

